PR #19, opened 2022-04-05
remove snapp from docs

iregina:


iregina:
Merging my own PR since this is mainly copy changes

PR #26, opened 2022-09-19
Fixed links

jackryanservia:


jasongitmail:
Outdated. Jack pushed commits directly to another bigger branch. But we can manually check these items to confirm they're good

PR #29, opened 2022-09-20
Nodedevcontent

iregina:
Content for the node developers section


iregina:
Checks failed cause there are some dead/broken links which I removed temporarily

docs/node-developers/sandbox-node.mdx
```diff
@@ -0,0 +1,70 @@
...
+docker logs --follow mina
+```
+
+And stop mina by running.
+
+```
+docker stop mina
+```
+
+You can use the Mina CLI to interact with the sandbox node. The following command opens a shell inside the docker container from where you can issue any of the available [mina commands](/node-operators/cli-reference).
```
jasongitmail, line 46: Can we change this link text to `cli commands` please? I know it's old content, but it'll make more sense to readers.

jasongitmail (approved):
Could you add the `A new version of Mina Docs is coming soon! We're currently updating the contents of page.` note box above the H1 of each page please?

Left one small comment. Otherwise LGTM!

docs/node-developers/sandbox-node.mdx
iregina: Done

jasongitmail:
Mentioned in Slack yesterday. But to ensure it's not lost--we used `note` for those callouts previously. After that, LGTM

PR #34, opened 2022-09-21
unhide redirect links

iregina:


jasongitmail (approved):
Jack tested this. Merging so we can finish final review

PR #35, opened 2022-09-21
Feature: Important items to be added

MartinMinkov:
# Important 
- On mobile, make the background for “On this page” dropdown the default, grayish color again. The default some kind of light gray with rounded corners. Problem is that it’s not very usable without a background and this seems to have inherited something unintended.

- Update syntax highlighting colors, to be more similar to previously used, to fit better with brand colors.

- On small desktop and mobile, the homepage hero image is squished (stretched vertically for some reason)

- Let’s tweak sidebar’s line-height for situations when an items has text on two lines )e.g. zkApps for Ethereum Developers”, so that the line height is smaller. I.e. the old docs is easier to distinguish what is a multi-line item vs another single line item within the sidebar. Try `line-height:1.5rem`, that is what the old docs seems to have. Then `padding: 0.6rem 0;` is used to separate different rows of items.

- On mobile, make the background for “On this page” dropdown the default, grayish color again. The default some kind of light gray with rounded corners. Problem is that it’s not very usable without a background and this seems to have inherited something unintended.

# Nice To Have
- In sidebar, change the dropdown arrow’s rotation animation to be a little faster (so it feels more similar to our now, quicker dropdown speed which is great)

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -97,8 +97,7 @@ const katex = require("rehype-katex");
         },
 
         prism: {
-          theme: darkCodeTheme,
-          darkTheme: darkCodeTheme,
+          theme: require("prism-react-renderer/themes/oceanicNext"),
         },
 
         colorMode: {
--- a/src/components/features/HomepageFeatures.module.scss
+++ b/src/components/features/HomepageFeatures.module.scss
@@ -54,13 +54,8 @@
   justify-content: space-between;
 
   & > img {
-    width: 18rem;
-    height: 15rem;
-
-    @media screen and (min-width: variables.$screen-lg) {
-      width: 28rem;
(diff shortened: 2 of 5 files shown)


jasongitmail:
Looks really good! The sidebar is easy to read even for multi-line items now! And color syntax is much better!

- One regression, the currently-selected item in the left sidebar now has a background color, but shouldn't. (The new lighter background color itself which is used on _hover_ is great.)
- On mobile, could we make the background for "on this page" a bit darker please? So it's easier to see.  Should we try the dropdown style from the brand guidelines too (i.e. like used in the sidebar) on this? There is a risk it will become hard to see given it's thinness and small size, but maybe we should try?

jasongitmail:
Looks great! Thank you!

PR #61, opened 2022-10-09
Tutorial 03 - deploying to berkeley

es92:


docs/zkapps/tutorials/03-deploying-to-a-network.mdx
```diff
@@ -12,21 +19,563 @@ In previous tutorials, we've seen how to deploy and run transactions on a local
...
+ 22   );
+ 23   Mina.setActiveInstance(Berkeley);
+ 24
+ 25   let transactionFee = 100_000_000;
+ 26 
+ 27   // ----------------------------------------------------
+...
+ ```
+
+ While earlier we set the active instance to a local blockchain, we are now setting the active instance to the remote berkeley network. We are connecting to Berkeley through a graphql proxy, which is running a Mina node connected to the Berkeley network.
```
jasongitmail, line 211: berkeley network > Berkeley network
graphql > GraphQL

docs/zkapps/tutorials/03-deploying-to-a-network.mdx
```diff
@@ -12,21 +19,563 @@ In previous tutorials, we've seen how to deploy and run transactions on a local
...
+ 24
+ 25   let transactionFee = 100_000_000;
+ 26 
+ 27   // ----------------------------------------------------
+...
+ ```
+
+ While earlier we set the active instance to a local blockchain, we are now setting the active instance to the remote berkeley network. We are connecting to Berkeley through a graphql proxy, which is running a Mina node connected to the Berkeley network. 
+
+ You can also run a Mina node locally, and instead use its graphql endpoint. While in other blockchains this would be very heavyweight, because Mina is succinct this is actually a reasonable option. See [here](/node-operators/getting-started) to see how to do this.
```
jasongitmail, line 213: graphql > GraphQL

jasongitmail (commented):
I've reviewed to line 215 and will resume there

docs/zkapps/tutorials/03-deploying-to-a-network.mdx
```diff
@@ -12,21 +19,563 @@ In previous tutorials, we've seen how to deploy and run transactions on a local
...
+
+With the Ethereum Bridge (future tutorial coming), states and proofs can also be available to Ethereum and other EVM chains.
+
+ We also set a transaction fee, which we'll use to pay for access to sending transactions and deploying smart contracts on Mina. Transaction fees in code are inputted as nano mina. We will use a 0.1 Mina fee as default (100,000,000 nano mina)
+
+ ```ts
+ ...
+ 18   // ----------------------------------------------------
+ 19
+ 20   const Berkeley = Mina.BerkeleyQANet(
```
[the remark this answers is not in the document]
es92: gotcha, thanks good to know (and makes sense)

docs/zkapps/tutorials/03-deploying-to-a-network.mdx
```diff
@@ -1,32 +1,456 @@
...
+
+You can run this now with:
+
+```sh
+npm run build && node build/src/main.js berkeley
+```
+
+Which will read keys from `keys/berkeley.json`. Public and Private Keys in Mina are commonly stored in `Base58` for easily readability. In Mina, public keys start with `B62`, and private keys start with `EKE` for easy differentiability.
+
+Our Smart Contract is also stored on chain in the same account we deployed from. So we set `zkAppPrivateKey = deployerPrivateKey`.
```
jasongitmail, line 279: > Our Smart Contract is also stored on chain in the same account we deployed from

Problematic b/c the smart contract isn't stored on chain. 

Could this sentence be excluded? If not, maybe something along the lines of: "Our zkApp account address is equal to the public key for the key pair that we created when using the `zk config` command." Idk, would probably just exclude it

docs/zkapps/tutorials/03-deploying-to-a-network.mdx
```diff
@@ -1,32 +1,456 @@
...
+ 56
+ 57   console.log(
+ 58     `Using fee payer account with nonce ${account.nonce}, balance ${account.balance}`
+ 59   );
+ 60
+ 61   // ----------------------------------------------------
+ ...
+```
+
+ We wait until the deployer's account exist - if it does not, we share a link with the user to go to the faucet and request funds.
```
jasongitmail, line 317: The user is the developer we're talking to, so maybe we can tweak the wording.

docs/zkapps/tutorials/03-deploying-to-a-network.mdx
```diff
@@ -1,32 +1,456 @@
...
+
+If you've already run `zk deploy` as described earlier in this tutorial, then this should find the account instantly and move on to the next step. But its good practice to have as a check, and useful if you're writing code for other flows, say a CLI application that interacts with an existing zkapp, where the user's key may not have funds yet.
```
jasongitmail, line 319: Given this is only useful when scripting deployment, maybe this block of code should only exist in the utils related to deployment? Opportunity to simplify. Not a strong opinion

jasongitmail (commented):
Much leaner! Looks nice! 

I went through the text today. Tomorrow I'll go through and try the code out too. I still have mixed thoughts about the length of the 2nd portion, but this is much improved and its value will probably become clear to me after trying out all the code tomorrow.

bkase (commented):
Love this tutorial -- really straightforward 💯 

I am 100% sure that the "Interacting with our Deployed Smart Contract" distracts from the tutorial flow and this content needs to move somewhere else and can be linked to from this page but shouldn't be inlined -- keep in mind in the next tutorial or so folks will be interacting with their contracts from the UI and the browser wallet

es92:
Cool - wdyt if we move the "Interacting with our Deployed Smart Contract" section to a tutorial outside of the numbered sections? Happy to discuss more, but I think it will be really useful to people, even with the following tutorial showing how to write a UI, so would like it to live somewhere.

bkase:
agreed that it is useful and needs to live somewhere too, defer to @jasongitmail to find a good spot for it. We can always fallback to a off-docs-site temporarily as well (like a Gist or README or notion page)

jasongitmail (approved):
@bkase Evan mentioned something important today, which is his 2nd half of this also demonstrates how to call the `init()` method.

I've proposed a short & long term solution, which @es92 agreed with:
1. Short term: Merge tutorial as is, because it shows how to call the init method.
2. Long term--i.e. [after the deploy method calls `init()`](https://github.com/o1-labs/snarkyjs/issues/494) so this part of the DX is improved, then we move the content for the 2nd half of this into a separate tutorial, potentially with a title like "Interacting with zkApps Server-side" or similar.

es92:
One note on that, we probably still want to have some way to allow init functions that take an argument - but imo that can probably live in the "Interacting with zkApps Server-side" tutorial as a more advanced way of deploying things

PR #79, opened 2022-10-21
Fix: typos

omahs:
Fix: typos


ymekuria (approved):
Thanks for fixing this!

PR #88, opened 2022-10-30
Tutorial 5, Common Types and Functions

es92:


docs/zkapps/tutorials/05-common-types-and-functions.mdx
```diff
@@ -0,0 +1,358 @@
...
+points8 Fields: 0,0,1,10,2,20,3,30,4,40,5,50,6,60,7,70
+```
+
+A great example of a complex use of CircuitValue is the MerkleTree implementation  - see the source for that [here](https://github.com/o1-labs/snarkyjs/blob/main/src/lib/merkle_tree.ts) for more.
+
+## Control flow
+
+There are two functions which help do control flow within SnarkyJS:
+
+* [Circuit.if](http://localhost:3000/zkapps/snarkyjs-reference/classes/Circuit#if)
```
Trivo25, line 195: broken link

docs/zkapps/tutorials/05-common-types-and-functions.mdx
```diff
@@ -0,0 +1,358 @@
...
+* [Circuit.switch](http://localhost:3000/zkapps/snarkyjs-reference/classes/Circuit#switch)
```
Trivo25, line 196: broken link

sidebars.js
```diff
@@ -54,7 +54,7 @@ module.exports = {
           items: [
             "zkapps/tutorials/hello-world",
             "zkapps/tutorials/private-inputs-hash-functions",
-            "zkapps/tutorials/deploying-to-a-network",
+            "zkapps/tutorials/common-types-and-functions",
```
Trivo25, line 57: should this be replaced or rather be a new entry?

Trivo25 (changes requested):
Thanks! this will help users get up to speed :) 
I just left a few small comments :)

mitschabaude:
I think we should probably release the new Struct and refactor the CircuitValue parts of this tutorial to it (https://github.com/o1-labs/docs2/pull/88#discussion_r1017781852)

Since now with the imminent testnet redeployment we can do a new release

docs/zkapps/tutorials/05-common-types-and-functions.mdx
```diff
@@ -0,0 +1,358 @@
...
+
+  const point1 = Point.fromField(Field(10), Field(4));
+  const point2 = Point.fromField(Field(1), Field(2));
+
+  const pointSum = Point.addPoints(point1, point2);
+
+  console.log(`pointSum Fields: ${pointSum.toFields().map((p) => p.toString())}`);
+
+  class Points8 extends CircuitValue {
+    points: Point[];
```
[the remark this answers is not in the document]
mitschabaude: yeah please let's do that! the outside-decorator usage was just a hack for a case where decorators didn't work (Merkle tree IIRC)

docs/zkapps/tutorials/05-common-types-and-functions.mdx
```diff
@@ -0,0 +1,358 @@
...
+    // initialize the zkapp
+    const zkapp = new BasicMerkleTreeContract(basicTreeZkAppAddress);
+
+    // create a new tree
+    const height = 20;
+    const tree = new Experimental.MerkleTree(height);
+    class MerkleWitness extends Experimental.MerkleWitness(height) {}
+
+    // deploy the smart contract
+    const deploy_txn = await Mina.transaction(deployerAccount, () => {
```
mitschabaude, line 317: nit: I think we should stick to strict camelCase - `deployTxn`

docs/zkapps/tutorials/05-common-types-and-functions.mdx
```diff
@@ -0,0 +1,358 @@
...
+      AccountUpdate.fundNewAccount(deployerAccount);
+      zkapp.deploy({ zkappKey: basicTreeZkAppPrivateKey });
+      // get the root of the new tree to use as the initial tree root
+      zkapp.init(tree.getRoot());
+      zkapp.sign(basicTreeZkAppPrivateKey);
+    });
+    await deploy_txn.send().wait();
+
+    const incrementIndex = 522;
```
mitschabaude, line 326: nit: the code below would become slightly cleaner if we instantiate incrementIndex as a bigint literal here: 
```ts
const incrementIndex = 522n;
```

mitschabaude (approved):
Amazing tutorial @es92 - thanks for helping out with the docs so much! I just have a couple of small comments

docs/zkapps/tutorials/05-common-types-and-functions.mdx
```diff
@@ -0,0 +1,358 @@
...
+
+# Tutorial 5: Common Types and Functions
+
+Welcome to tutorial 5. In previous tutorials, we've seen how to deploy smart contracts to the network, and we've interacted with them from both a React UI, nodejs, and the CLI.
+
+In this tutorial, we will talk about more types that are useful when building with SnarkyJS, so you can build more applications. So far, we've mostly been using the `Field` type. While in theory you could do everything with just Fields, SnarkyJS provides many more useful types to help with development.
+
+You can find all of these on the [SnarkyJS Reference page](../snarkyjs-reference), for their full API documentation.
+
+There is also a project [here](https://github.com/es92/zkApp-examples/tree/main/05-common-types-and-functions/src), with a [main.ts](https://github.com/es92/zkApp-examples/blob/main/05-common-types-and-functions/src/main.ts) demoing the concepts presented in this tutorial, along with smart contracts showing more advanced usage of some of the concepts, particularly Merkle Trees.
```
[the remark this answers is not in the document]
es92: Yeah agree

sidebars.js
Trivo25: ahh okay, I see!

docs/zkapps/tutorials/05-common-types-and-functions.mdx
es92: We currently target ES2019 with zk project - so the bignum notation doesn't work unfortunately - do think we should consider updating that tho to unlock newer syntax like this

mitschabaude: oh I see -- yes I'll make a note to update the target!

docs/zkapps/tutorials/05-common-types-and-functions.mdx
```diff
@@ -0,0 +1,360 @@
...
+Fields in signature: 256
+```
+
+## CircuitValue
+
+:::info
+`CircuitValue` is soon to be replaced with `Struct` in an upcoming version of SnarkyJS. Expect this tutorial to update to use `Struct` in the future accordingly.
+:::
+
+A special type is [CircuitValue](../snarkyjs-reference/classes/CircuitValue), which lets you create your own types for your application.
```
jasongitmail, line 142: > [CircuitValue](../snarkyjs-reference/classes/CircuitValue) allows you to create your own compound data types.

Preference for clarity

docs/zkapps/tutorials/05-common-types-and-functions.mdx
```diff
@@ -0,0 +1,360 @@
...
+
+All arguments passed into smart contracts need to be arguments SnarkyJS can understand. Creating a CircuitValue allows you pass in structured pieces of data, which via their class definition can come with methods to manipulate them.
```
jasongitmail, line 144: Suggested rephrasing: 
> Your `CircuitValue` can be defined as one or more data types that SnarkyJS understands--i.e. Field, higher-order types built into SnarkyJS based on Field, or other `CircuitValue` types defined by you. You can also define methods on your `CircuitValue` to act upon this data type, if desired, but doing so is optional.

docs/zkapps/tutorials/05-common-types-and-functions.mdx
```diff
@@ -0,0 +1,360 @@
...
+```
+
+Which prints the following to the console:
+
+```sh
+pointSum Fields: 11,6
+points8 Fields: {"points":[{"x":"0","y":"0"},{"x":"1","y":"10"},{"x":"2","y":"20"},{"x":"3","y":"30"},{"x":"4","y":"40"},{"x":"5","y":"50"},{"x":"6","y":"60"},{"x":"7","y":"70"}]}
+```
+
+A great example of a complex use of CircuitValue is the MerkleTree implementation  - see the source for that [here](https://github.com/o1-labs/snarkyjs/blob/main/src/lib/merkle_tree.ts) for more.
```
jasongitmail, line 193: backticks for `CircuitValue`

docs/zkapps/tutorials/05-common-types-and-functions.mdx
```diff
@@ -0,0 +1,360 @@
...
+
+## Control flow
+
+There are two functions which help do control flow within SnarkyJS:
```
jasongitmail, line 197: I think it'd be useful to start this paragraph explaining:

> Because the underlying proof system used by SnarkyJS--Kimchi--does not yet support branching, we cannot use regular `if` statements in our code. But we have two other, circuit-compatible options available to us for flow control:

docs/zkapps/tutorials/05-common-types-and-functions.mdx
```diff
@@ -0,0 +1,360 @@
...
+
+* [Circuit.if](../snarkyjs-reference/classes/Circuit#if)
+* [Circuit.switch](../snarkyjs-reference/classes/Circuit#switch)
+
+With these, you can write conditionals inside SnarkyJS.
```
jasongitmail, line 202: Suggestion:
> `Circuit.if` is similar to a ternary in JavaScript. `Circuit.switch` is similar to a switch case statement in JavaScript. With these, you can write conditionals using SnarkyJS.

(We've seen people misunderstand how to use `Circuit.if`, to attempt complicated branching logic, but we've found people grasp it easily when described as similar to a ternary.)

docs/zkapps/tutorials/05-common-types-and-functions.mdx
```diff
@@ -0,0 +1,360 @@
...
+inputSum: -5
+inputSumAbs: 5
+largest: 22
+```
+
+By inserting function calls, you can write circuits which usual if / else statements:
+
+```ts
+const result = Circuit.if(
+  <condition>,
```
jasongitmail, line 240: This might be easier to read if it avoided non-valid syntax and relied on a basic example (like a equals b) and comments for things like `// condition`

Otherwise I stopped to think, is this TS syntax, and then realized I just needed to mentally parse it.

docs/zkapps/tutorials/05-common-types-and-functions.mdx
```diff
@@ -0,0 +1,360 @@
...
+);
+```
+
+Note that unlike normal programming, because SnarkyJS under the hood is creating a zk cirucit, this is in fact executing both branches and selecting the one that matches the condition.
+
+## Assertions
+
+We've been using assertions throughout these tutorials already - but to clarify what they're doing, assertions are creating constraints in our circuit, that must hold true for a valid proof to be generated.
+
+Besides the `assertEquals` we've often used, there is also `assertTrue` available on the Bool class.
```
jasongitmail, line 258: These would be more easily understood as method names if they contain parenthesis:
`assertEquals()` and `assertTrue()`. Then any lacking parenthesis are more easily interpreted as property names, class names, etc.

docs/zkapps/tutorials/05-common-types-and-functions.mdx
```diff
@@ -0,0 +1,360 @@
...
+    const witness = new MerkleWitness(tree.getWitness(BigInt(incrementIndex)));
+
+    // update the leaf locally
+    tree.setLeaf(BigInt(incrementIndex), incrementAmount);
+
+    // update the smart contract
+    const txn1 = await Mina.transaction(deployerAccount, () => {
+      zkapp.update(
+        witness,
+        Field.zero, // leafs in new trees start at a state of 0
```
jasongitmail, line 343: `Field.zero` > `Field(0)`. The former is now deprecated

docs/zkapps/tutorials/05-common-types-and-functions.mdx
```diff
@@ -0,0 +1,360 @@
...
+Merkle trees are currently in the `Experimental` module of SnarkyJS. So start off by importing `Experimental`:
+```ts
+import {
+  ...
+  Experimental,
+  ...
+} from 'snarkyjs'
+```
+
+When using Merkle trees, you initialize an instance of the data structure on the "ui" side of your application at a particular height.
```
jasongitmail, line 273: As a reader, this paragraph leaves me like "what i don't get it...and there's no example". Unless I really understand Merkle trees already, I'm not going to get it.

Maybe instead, this could start event higher level, or move this paragraph before the 2nd example, which I assume it's more related to.

This and the following paragraph feel like an opportunity to go more ELI5. @Trivo25 wrote the Merkle tree docs and might be able to help with a suggestion

jasongitmail (commented):
Reviewed all except Merkle tee

jasongitmail:
tiny nit: It might be good to use  `===` in your sh output. Because in JS `==` will do type conversion before checking equality, and `===` will not, and the latter is more representative of how `.equals()` works.

docs/zkapps/tutorials/05-common-types-and-functions.mdx
es92: Do we think it ever could support branching? I thought it had to be fixed circuits, cool if this is possible tho

docs/zkapps/tutorials/05-common-types-and-functions.mdx
es92: Updated - lmk what you think

docs/zkapps/tutorials/05-common-types-and-functions.mdx
```diff
@@ -0,0 +1,360 @@
...
+## More Advanced Types
+
+4 more advanced types are:
+
+* [CircuitString](../snarkyjs-reference/classes/CircuitString)
+* [Private keys](../snarkyjs-reference/classes/PrivateKey)
+* [Public keys](../snarkyjs-reference/classes/Types.PublicKey)
+* [Signatures](../snarkyjs-reference/classes/Signature)
+
+One special thing to note, is that the default `CircuitString` has a max length of 128 characters. We'll see in the following section how to create custom types, where you could for example build your own strings, modifed to whatever length you would like.
```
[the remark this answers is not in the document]
mitschabaude: Hey @es92 @jasongitmail, sorry for unresolving this, but I'm not a huge fan of this formulation:
> SnarkyJS supports fixed length strings currently. Support for dynamic length strings will be added in the future after dynamic array access is available.

The CircuitString is "fixed _max_ length", not fixed length, and that makes a huge difference to the developer. The APIs on CircuitString feel like if the string were dynamic-length. The only thing that will change later is the performance (constraint count) of some of the string operations, not the API. (I'm also not sure whether we can ever get rid of the static max length, we'll see.)

So, I'd like us to present this, and the "static circuit" aspects in general, in a more favorable light. Suggest the following reformulation:
>  Under the hood, SnarkyJS strings are fixed size, but the API abstracts away the fixed length and makes them feel like normal, dynamic strings. This approach can have a high cost in terms of constraint count, i.e. proving performance. In the future, support for dynamic array access in our proof system will improve performance of `CircuitString` and potentially remove the limitation of a fixed maximum size.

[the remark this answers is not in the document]
mitschabaude: If we want to keep it simple for a tutorial, I actually suggest to only use the first sentence:

> Under the hood, SnarkyJS strings are fixed size, but the API abstracts this away and makes them feel like normal, dynamic strings.

I really don't think the performance disclaimer is important here. In other places, we show usage of comparison operators on `Field` (lower than, greater than) which are crazy inefficient atm, without a disclaimer.

docs/zkapps/tutorials/05-common-types-and-functions.mdx
jasongitmail: My understanding from Izzy a while back is that yes, he thinks there is a way to support true branching someday

docs/zkapps/tutorials/05-common-types-and-functions.mdx
```diff
@@ -0,0 +1,376 @@
...
+```
+
+Merkle Trees can be created in your application as so:
+
+```ts
+const height = 20;
+const tree = new Experimental.MerkleTree(height);
+```
+
+The height variable determines how many leaves are available to your application. A height of 20 for example will lead to a tree with `2**(20-1)`, or 524,288 leaves.
```
jasongitmail, line 287: Would `2^(20-1)` be more clear?

docs/zkapps/tutorials/05-common-types-and-functions.mdx
```diff
@@ -0,0 +1,376 @@
...
+
+Merkle trees in smart contracts are stored as the hash of the merkle tree's root. Smart contract methods which update the merkle root can take a "witness" of the change as an argument, which represents the merkle path to the data for which inclusion is being proved for.
```
jasongitmail, line 289: Can remove the `for` since you already have `for which`

docs/zkapps/tutorials/05-common-types-and-functions.mdx
```diff
@@ -0,0 +1,376 @@
...
+      zkapp.sign(basicTreeZkAppPrivateKey);
+    });
+    await txn1.send().wait();
+
+    // compare the root of the smart contract tree to our local tree
+    console.log(`BasicMerkleTree: local tree root hash after send1: ${tree.getRoot()}`);
+    console.log(`BasicMerkleTree: smart contract root hash after send1: ${zkapp.treeRoot.get()}`);
+```
+
+While in this example leafs are fields, you can put more variables in a leaf by instead hashing an array of fields, and setting a leaf to that.
```
jasongitmail, line 368: `leaves`

docs/zkapps/tutorials/05-common-types-and-functions.mdx
```diff
@@ -0,0 +1,376 @@
...
+```ts
+const height = 20;
+const tree = new Experimental.MerkleTree(height);
+```
+
+The height variable determines how many leaves are available to your application. A height of 20 for example will lead to a tree with `2**(20-1)`, or 524,288 leaves.
+
+Merkle trees in smart contracts are stored as the hash of the merkle tree's root. Smart contract methods which update the merkle root can take a "witness" of the change as an argument, which represents the merkle path to the data for which inclusion is being proved for.
+
+Here is a simple example of a contract that stores the root of a merkle tree, where each leaf stores a number, and the smart contract has an `update` function that adds a number to the leaf, while checking that the number added was less than 10:
```
jasongitmail, line 291: Suggestion to emphasize that this aspect is just for the example:

> , while requiring that the number added is less than 10, for this example

jasongitmail (approved):
@es92 The Merkle tree section looks great! Will be easy to understand. Thanks for updating!

jasongitmail (approved):

PR #95, opened 2022-11-01
Hello World Changes

duncand0nuts:
This PR has 3 commits: 

1. [add tip saying use copy button for code snippets](https://github.com/o1-labs/docs2/commit/a246fd73063f4df7f26ee6e80d499084757e9974) - advises users to use copy button to avoid copying line numbers inadvertently 
2. [change "contract" instance name to "zkAppInstance" to match source file](https://github.com/o1-labs/docs2/commit/d62cd2f6c5097eedb3a532a750b32f1e58ffc751) - change name of instance of `Square` to match the source file
3. [add info admonition telling users build commands will fail before Squ…](https://github.com/o1-labs/docs2/commit/13e97827cf933c91d4b1c350cb9222526214aea3) - warning box which tells users to expect build commands to fail if they run them when they are first mentioned

### DIFF
--- a/docs/zkapps/tutorials/01-hello-world.mdx
+++ b/docs/zkapps/tutorials/01-hello-world.mdx
@@ -326,17 +326,17 @@ Now, let's initalize our smart contract. Comments are provided to break down eac
  25   const zkAppAddress = zkAppPrivateKey.toPublicKey();
  26
  27   // Create an instance of our Square smart contract and deploy it to zkAppAddress
- 28   const contract = new Square(zkAppAddress);
+ 28   const zkAppInstance = new Square(zkAppAddress);
  29   const deployTxn = await Mina.transaction(deployerAccount, () => {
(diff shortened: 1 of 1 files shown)


es92 (approved):

jasongitmail (changes requested):
Thanks! I like the changes, except #2.

It's a good eye on the consistency. Thanks for catching that.

But imo we should change the source code to use `contract`, rather than vice versa. If you look at this code [side by side](https://github.com/o1-labs/docs2/pull/95/files#diff-539c8a54f551120053f57ce2ce398d7c7e2e87152c069ff12a2b88058251e069L323), one is visually easier to understand and takes less mental effort to understand what it's trying to represent, i.e. it's easier / more intuitive naming, so should be preferred.

jasongitmail (approved):
Thanks @duncand0nuts!

PR #106, opened 2022-11-11
Fix copy code snippet to not remove spread operators

### LINKED ISSUE #105: Fix copy code JS, to only remove ellipsis at the start of the line
Otherwise, it will remove the spread operator from within code too. I.e. `...Permissions.default(),` becomes `Permissions.default(),`.

ymekuria:
Fixes #105

### DIFF
--- a/src/theme/CodeBlock/CopyButton/index.js
+++ b/src/theme/CodeBlock/CopyButton/index.js
@@ -9,7 +9,7 @@ export default function CopyButton({ code, className }) {
   const copyTimeout = useRef(undefined);
   const handleCopyCode = useCallback(() => {
     const removeLineNumbers = /^\s*\d+\s/gm;
-    const removeEllipses = /^\.{3}/gm;
+    const removeEllipses = /(^\s*\.{3})(?![^a-z])|(^\.{3})/gm;
 
     let copiedCode = code.replace(removeLineNumbers, '');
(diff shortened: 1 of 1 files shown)


src/theme/CodeBlock/CopyButton/index.js
```diff
@@ -9,16 +9,21 @@ export default function CopyButton({ code, className }) {
   const copyTimeout = useRef(undefined);
   const handleCopyCode = useCallback(() => {
     const removeLineNumbers = /^\s*\d+\s/gm;
-    const removeEllipses = /\.{3}/gm;
+    const removeEllipses = /(^\s*\.{3})(?![^a-z])|(^\.{3})/gm;
```
jasongitmail, line 12: @ymekuria What's the first part for? I'd have thought your second part would solve it `(^\.{3})`

jasongitmail (approved):
Tested and works well. Thanks!

src/theme/CodeBlock/CopyButton/index.js
ymekuria: In testing I found that sometimes the elipses have preceding spaces. The first part ensures that spread operators are not removed when matching elipses that have preceding spaces.

jasongitmail: cool. thanks for explaining

PR #122, opened 2022-11-16
Init() and other 0.7.0 updates

mitschabaude:
Document `init()` and `Circuit.log()`, remove `Experimental` in some places, fix some code that was broken by API changes

### DIFF
--- a/docs/zkapps/how-to-write-a-zkapp.mdx
+++ b/docs/zkapps/how-to-write-a-zkapp.mdx
@@ -473,31 +473,27 @@ However, the secret itself remains private, because it can't be deduced from it'
 
 ##### Initializing state
 
-We didn't cover how to initialize on-chain state yet. This can be done in the `deploy()` method.
+We didn't cover how to initialize on-chain state yet. This can be done in the `init()` method.
 
-Like the constructor, `deploy()` is predefined on the base `SmartContract` class.
-It will be called when you deploy your zkApp with the zkApp CLI.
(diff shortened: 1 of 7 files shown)


docs/zkapps/how-to-write-a-zkapp.mdx
```diff
@@ -282,6 +279,28 @@ Under the hood, every `@method` defines a zk-SNARK circuit. From the cryptograph
 
 :::
 
+<!-- TODO: create at least a basic section about zk circuits below this subsection which contains (among other things) a more thourough explanation of what this tip hints at -->
+
+:::tip
+
+You will find that inside a `@nethod`, things sometimes behave a little differently. For example, the following code can't be used in a method where `x: Field` is an input parameter:
```
jasongitmail, line 286: `@method` (typo)

jasongitmail, line 286: Would suggest we reword this section to lead with the most important thing, and then follow with the explanation. It will be shorter and they'll know what the entire section is about. But let's skip that today, to save time.

jasongitmail (approved):
Thanks! LGTM. One important typo, I'll submit a commit, and then merge

docs/zkapps/how-to-write-a-zkapp.mdx
jasongitmail: Fixed

PR #125, opened 2022-11-17
Split out server side instructions from tutorial 3 into a separate tutorial

es92:


jasongitmail (approved):
Approving to clean up 3. We can have something dig into further review of the Interacting with a zkApp Server Side content at a later time

PR #135, opened 2022-11-21
remove Auro version info box

duncand0nuts:
v2.1.2 is live in Chrome store now


ymekuria (approved):
Glad we could remove this! LGTM

PR #160, opened 2022-11-23
Update tutorial 01 and 02 sign call

SebastienGllmt:
The tutorial 1 still uses the deprecated sign method. This PR updates it to use the new `requireSignature` approach

This should also be updated in the zkApp-examples repo, but I noticed this PR follows a slightly different approach, so it may be better to unify to it: https://github.com/es92/zkApp-examples/pull/5


barriebyron:
Thank you @SebastienGllmt  for the update, apologies for the very delayed response.
The docs and examples have been updated meanwhile, so I'll close this PR.
We appreciate your contributions!

PR #165, opened 2022-11-25
Tutorial 05: Add hints for console output

SebastienGllmt:
The console log outputs aren't explained. We should at least have some hints about why this is the output

Ideally we would have links to learn more, but I've been told the [signature description documentation](https://github.com/MinaProtocol/mina/blob/develop/docs/specs/signatures/description.md#signature-generation) is incorrect because "snarkyjs is not performing the algorithm from that doc, it's much simpler because it uses a random scalar for the nonce instead of deriving it"


SebastienGllmt:
Yeah, I have no strong preference on where the information is, but I think it would be useful for the section that explains the PrivateKey and PublicKey type to mention how these types are actually encoded. It could be a comment, or a paragraph that explains the output

barriebyron:
take a look now @mitschabaude  wdyt?

docs/zkapps/tutorials/05-common-types-and-functions.mdx
```diff
@@ -161,6 +161,12 @@ signature verified for data2: true
 Fields in signature: 256
 ```
 
+where:
+
+- `Fields in private key: 255` is a bit field element of the scalar field of the curve
```
mitschabaude, line 166: I don't think "bit field" is a good term, to me it's confusing. So, to me, this doesn't clearly explain why a private key has 255 field elements, and I'm also questioning whether this is a good place to explain that in the first place.

mitschabaude (commented):
Maybe this could be turned into a useful footnote, but as is I think it just hurts the text

docs/zkapps/tutorials/05-common-types-and-functions.mdx
barriebyron: @mitschabaude shall we forgo comments altogether? Or do you have a better way to describe?

barriebyron:
gist: document the PrivateKey and PublicKey type to mention how these types are actually encoded, useful for console log output description

PR #167, opened 2022-11-28
Document how SnarkyJS creates constraints, as a FAQ

### LINKED ISSUE #109: Describe how snarkyjs code creates constraints while executed
The dominant paradigm in today's zk tooling is that some user code is _compiled_ to a zk circuit. SnarkyJS, however, works differently: The user-written JS is _executed_ to create a circuit, not compiled.

This difference creates a lot of misconceptions about snarkyjs, for users coming from other zk ecosystems or from higher-level zk material, so we should call it out in the docs to prevent these misconceptions early on.

mitschabaude:
- closes #109 
- adds FAQ page

### DIFF
--- a/docs/zkapps/faq.mdx
+++ b/docs/zkapps/faq.mdx
@@ -2,4 +2,43 @@
 title: FAQ
 ---
 
-Ask a question (link to github or where to submit Q?)
+## Does SnarkyJS compile my JavaScript code to an arithmetic circuit?
+
+No, SnarkyJS **does NOT compile JS into anything else**! This is a very common misconception about how SnarkyJS works. It may be caused by the emphasis on compilation in other zk ecosystems -- many of which have custom languages for writing zk circuits.
+
+SnarkyJS, by contrast, is just a JavaScript library. It creates zk circuits from user code by _executing_ that code. For example, if you have a SmartContract with a `@method myMethod()`, during proof generation we will simply call that method: `myMethod();`
+
+This works because SnarkyJS sets up some global state - a "circuit" - where it collects variables and constraints, and functions like `Field.mul` or `Bool.assertEquals`, which you use inside your method, add variables/constraints to that global circuit.
+
+This has some implications for how your code behaves, and what part of it is included in the zk proof:
+
+- The way to turn your logic into a proof is to use SnarkyJS built-in datatypes, like `Field`, and SnarkyJS functions that operate on them, like `Field.mul()`.
+  - For example, `x.mul(y)` will add a generic PLONK gate to your circuit. It will also return a variable that you can use in further statements, which will get wired to the mutliplication gate.
+  - There are a few SnarkyJS functions which allow you to turn normal JS datatypes into Field elements and back - for example, `Field.toString()`. These functions don't add anything to your circuit. They will typically have a doc-comment clarifying this.
+- If you do normal JS stuff, without using SnarkyJS built-ins, like `'hello world'.split('').join(' ')`, none of that will add anything to your circuit, and so a statement like this won't be included in your zk proof in any way.
+  - Why? Because it doesn't call any of the functions that add stuff to the circuit
+  - There's nothing wrong with doing circuit-unrelated stuff like the above inside your method, as long as you're aware of what it's (not) doing.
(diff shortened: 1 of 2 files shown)


bkase (approved):
I think this is as concise as we can get, every point here is important to include imo. My recommendation would be to migrate this to its own section of some kind of "advanced internals" documentation (whenever we create that) and then reference with a backlink to that section from the FAQ.

For now, given a lack of anywhere else to put it, I think this is a good place.

Also, I think the content itself is great!

mitschabaude:
> My recommendation would be to migrate this to its own section of some kind of "advanced internals" documentation (whenever we create that) and then reference with a backlink to that section from the FAQ.

+1, we put this in a FAQ entry as a temp solution. Most of it can (should) live in an in-depth "SnarkyJS user manual" that some users have been asking for. Also, I had planned to link to my circuit-writing tutorial from here

docs/zkapps/faq.mdx
```diff
@@ -4,21 +4,20 @@ title: FAQ
...
+This has some implications:
 
-This has some implications for how your code behaves, and what part of it is included in the zk proof:
+- In order to turn your logic into a proof, you'll need to use SnarkyJS built-in datatypes such as `Field`. You'll also need to use the SnarkyJS functions that operate on them, like `Field.mul()`.
+  - A statement like `x.mul(y)` will add a generic PLONK gate to your circuit. It will also return a variable that you can use in further statements, which will get wired to the multiplication gate.
+  - Some SnarkyJS functions allow you to turn normal JS datatypes into `Field` elements and back, such as `Field.toString()`. These functions don't add anything to your circuit. They will typically clarify this in a doc-comment.
+- Conventional JS code such as `'hello world'.split('').join(' ')` that doesn't use SnarkyJS built-ins will not be included in your zk proof in any way since it will not add anything to your circuit.
+  - Why? Because it doesn't call any of the functions that build the circuit.
+  - There's nothing wrong with having circuit-less code like the above inside your method, as long as you're aware of what it's (not) doing.
```
mitschabaude, line 18: > circuit-less

not sure if I like "circuit-less code" here better than "circuit-unrelated code" 🤔 maybe "non-circuit code"?

docs/zkapps/faq.mdx
```diff
@@ -27,18 +26,16 @@ for (let y of [5, 10, 15]) {
...
+This example fails for two reasons:
 
-1. Fundamentally, your methods must be written such that they create the same constraints every time. A proof cannot be verified against the verification key for a different set of constraints. The code above, however, tried to add an `assertEquals` contraint depending on the value of `b`.
-2. More practically, you can't read out the values of variables in circuit code - an expression like `b.toBoolean()` will throw an error during `SmartContract.compile()` (which generates the prover and verification keys). This is because during `compile()`, variables like `b` don't have any JS values attached to them; they represent abstract variables used to build up an abstract arithmetic circuit.
+1. Circuit code has to interact with SnarkyJS built-in types. An expression like `b.toBoolean()` will throw an error during `SmartContract.compile()` (which generates the prover and verification keys) because you can't read out the values of variables like that. This is because during `compile()`, variables like `b` don't have any JS values attached to them; they represent abstract variables used to build up an abstract arithmetic circuit.
```
mitschabaude, line 38: I'm not sure the first sentence "Circuit code has to interact with SnarkyJS built-in types." does a good job of summarizing this bullet point.

I'd consider the following:

> 1. `b.toBoolean()` can't be used in circuit code at all. It will throw an error during `SmartContract.compile()` because ...

docs/zkapps/faq.mdx
```diff
@@ -27,18 +26,16 @@ for (let y of [5, 10, 15]) {
...
+2. More subtly, your methods must create the same constraints every time because a proof cannot be verified against a verification key for a differing set of constraints. The code above adds an `assertEquals` _on condition of_ the value of `b`. This can lead to constraints varying between executions of the proof.
 
-Takeaway: SnarkyJS metaprogramming only works if it doesn't try to interact with the concrete program that's being written.
+Takeaway: Conventional JS constructs only work if they don't interact with the concrete zk logic being written in SnarkyJS.
```
mitschabaude, line 41: I guess this last statement wasn't very precise to begin with, and became even more vague, I'll remove it

docs/zkapps/faq.mdx
nicc: Cool, yes I agree and I like your edit

jasongitmail (approved):
@mitschabaude added a comment, but merging

PR #170, opened 2022-11-29
Add a CI job to auto build SnarkyJS API docs

MartinMinkov:
## Description
This PR adds a Github Actions job that builds new SnarkyJS API reference docs and copies them over to the docs repo to be pushed and updated. This can be triggered manually with `workflow_dispatch`.

It starts by checking out both the SnarkyJS repo and the docs repo, installing dependencies, running `typedoc-plugin-markdown`, copying over the generated reference docs into the docs repo, updating and formatting `sidebars.js`, and finally configuring git and pushing to the docs repo.

## Tested
This was tested manually on my system by cloning both SnarkyJS and docs2 and running through the commands manually up to the point of pushing on git.

### DIFF
--- a/.github/workflows/snarkyjs-api-reference.yml
+++ b/.github/workflows/snarkyjs-api-reference.yml
@@ -1,0 +1,33 @@
+on:
+  workflow_dispatch:
+    inputs:
+      logLevel:
+        description: 'Log level'
+        required: true
+        default: 'warning'
+        type: choice
+        options:
+          - info
+          - warning
+          - debug
+      tags:
+        description: 'Test scenario tags'
+        required: false
+        type: boolean
+      environment:
+        description: 'Environment to run tests against'
+        type: environment
+        required: true
+
+jobs:
+  log-the-inputs:
+    runs-on: ubuntu-latest
+    steps:
+      - run: |
+          echo "Log level: $LEVEL"
+          echo "Tags: $TAGS"
+          echo "Environment: $ENVIRONMENT"
+        env:
+          LEVEL: ${{ inputs.logLevel }}
+          TAGS: ${{ inputs.tags }}
+          ENVIRONMENT: ${{ inputs.environment }}
--- a/.github/workflows/snarkyjs-api-reference.yml
+++ b/.github/workflows/snarkyjs-api-reference.yml
@@ -1,12 +1,6 @@
 name: SnarkyJS-API-Reference
-on:
-  workflow_dispatch:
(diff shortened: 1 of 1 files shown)


mitschabaude (approved):
lgtm!

Trivo25 (approved):

.github/workflows/snarkyjs-api-reference.yml
```diff
@@ -0,0 +1,56 @@
+name: SnarkyJS-API-Reference
+on:
+  workflow_dispatch:
+    snarkyjs_branch:
+      type: choice
+      description: SnarkyJS branch to build from
```
jasongitmail, line 6: Cool, I didn't know 'choice' was an option in Github Actions.

What is the benefit you envision here? Wouldn't we want to always consistently follow whichever branch matches the published version of SnarkyJS (i.e. `releases` curretly)?

.github/workflows/snarkyjs-api-reference.yml
```diff
@@ -0,0 +1,56 @@
...
+          # Update with new enteries and format sidebars.js
+          node update-sidebars-snarkyjs-api.js && npx prettier --config .prettierrc --write docusaurus.config.js
+
+          # Configure git and push
+          git config user.name github-actions
+          git config user.email github-actions@github.com
+          git add docs/zkapps/snarkyjs-reference
+          git add docusaurus.config.js
+          git commit -m "Build and publish new SnarkyJS API reference docs"
+          git push
```
jasongitmail, line 56: Any benefit to PR'ing it in stead of pushing directly? This current way is more similar to how it will need to be set up for the eventually fully automated solution, assuming it works with branch protection rules. Lmk if we need to tweak those.

jasongitmail (approved):
LGTM! Excited for this! Left a couple questions, but it's good as is.

.github/workflows/snarkyjs-api-reference.yml
MartinMinkov: That's a good point! I suppose the benefit of creating a PR would be that you can create a deploy link and inspect changes if you wish. I will change it to create a PR instead :D

.github/workflows/snarkyjs-api-reference.yml
MartinMinkov: At the time, I figured that it might be useful if we made a documentation change in SnarkyJS but the more I think about it, I think it should just be one branch (releases) as you suggest. I'll fix this in the PR.

PR #171, opened 2022-11-29
fix typeof state storage field

aodhgan:


mitschabaude:
this looks wrong to me! what was the motivation for this change?

mitschabaude (changes requested):
probably shouldn't be merged

aodhgan:
The motivation is that the example code will not compile like that - the change in this commit was wrong tho. I have opened https://github.com/o1-labs/docs2/pull/175 instead.

PR #176, opened 2022-11-30
Docs comparing actions/events

bkase:


mitschabaude (approved):
Very useful, thanks!

mitschabaude:
Something else that came up in the last builder's call: We should state that Events / Actions are public, not private

PR #186, opened 2022-12-07
Fix import for "CreditScoreOracle.test.ts"

0xkeivin:
import function `OracleExample` for the test file "CreditScoreOracle.test.ts" should be from `./CreditScoreOracle` and not `./OracleExample` as per beginning of the guide. 
With this change, it passes the `npm run test` command


mitschabaude (approved):
thanks! cc @jasongitmail

PR #189, opened 2022-12-12
Update graphql-api.mdx

Isaac-DeFrain:
Fix stale link


barriebyron:
Hi @Isaac-DeFrain  thanks for this PR and apologies for the very delayed response.
the archive schema is distinct from the GraphQL schema, so this doesn't work

PR #197, opened 2022-12-15
[updated]Broken Link

nikbhintade:
The link in the `What are zero-knowledge proofs?` was broken due to incorrect syntax just changed that to correct syntax.

Link to the page where link was broken [https://docs.minaprotocol.com/about-mina](https://docs.minaprotocol.com/about-mina)

Incorrect syntax:
image


mitschabaude (approved):
thanks! cc @jasongitmail

PR #202, opened 2022-12-19
Add edit page button

ymekuria:
Addresses #80 

This PR adds a custom `edit this page` button above the generated right page table of contents sidebar. The button is still rendered in the case when the right sidebar is not generated.


jasongitmail (changes requested):
@ymekuria A few changes

-  Can we hide this button on the homepage please?
-  Does button style match brand design guidelines? (e.g. I notice the submit button in the footer has white text for a similar button. Similarly, on hover, the style feels non-standard with the orange text. Worth checking.) ... actually the orange is bright to have on every page; maybe we should duplicate the black button style on the homepage style to try here too--wdyt?
-  Could you add some margin below it please, so it the right sidebar doesn't bump into it.

src/theme/DocItem/Layout/styles.module.css
```diff
@@ -0,0 +1,10 @@
+.docItemContainer header + *,
```
MartinMinkov, line 1: Can we use `scss` for this file instead of `css`? We should be using that by default since it's used everywhere else.

MartinMinkov (commented):
The `Edit this page` label that Docusaurus defaults too is showing on each page in addition to the button so now we have two places to edit content :confused: Mind taking a look at seeing how that can be fixed?

MartinMinkov (commented):
On mobile, the button is shown below all the markdown content and navigation buttons. Not sure if this is intended but I think it should be at the top, like on the desktop as well! :D

ymekuria:
> The `Edit this page` label that Docusaurus defaults too is showing on each page in addition to the button so now we have two places to edit content 😕 Mind taking a look at seeing how that can be fixed?

I can figure out how to remove it. I left it at the bottom as a redundancy after discussing with Damien. @jasongitmail @MartinMinkov do you both prefer removing the redundant `edit this page` link at the bottom of every page?

ymekuria:
> * Can we hide this button on the homepage please?
@jasongitmail The original `edit page link` docusuraus generates is on the homepage. There could be error and typos on the homepage that someone would want to make a pr to fix. I'm happy to remove it if you prefer. 
> * Does button style match brand design guidelines? (e.g. I notice the submit button in the footer has white text for a similar button. Similarly, on hover, the style feels non-standard with the orange text. Worth checking.) ... actually the orange is bright to have on every page; maybe we should duplicate the black button style on the homepage style to try here too--wdyt?
I agree the orange is a little jarring. I'll update with the dark style so we can see what it looks like.
> * Could you add some margin below it please, so it the right sidebar doesn't bump into it.
100% I meant to add margin as well as fix the text and hover styles you mentioned.

jasongitmail:
Seems fine to remove the default edit link +1 But not a big deal if it's difficult.

ymekuria:
> Seems fine to remove the default edit link +1 But not a big deal if it's difficult.

I removed the edit link and made the changes. Let me know what you think about the dark theme button.

jasongitmail:
This looks great with recent changes! 

Suggesting two positioning tweaks for design:
- Changes to margin-left and margin-bottom for the edit button container:
  Screenshot 2022-12-22 at 7 43 04 PM
- And increasing (4rem > 6rem looks good) spacing of the TOC from this button when scrolled, so breathing space remains:
  Screenshot 2022-12-22 at 7 42 26 PM

Noticed the button is hidden [on this one page](https://docs2-git-feature-edit-page-button-minadocs.vercel.app/about-mina/what-are-zero-knowledge-proofs ), for some reason. Not sure if that's intentional for your testing purposes.

When scrolled to the bottom of the page, the two sticky elements scroll off at different times. Possibly to tweak this to make it happen cleanly?:
Screenshot 2022-12-22 at 7 48 39 PM

jasongitmail (approved):
Looks great @ymekuria !

I just realized, on https://docs2-pflexbr03-minadocs.vercel.app/about-mina/what-are-zero-knowledge-proofs, it must not create a sidebar because there are no sub-heads

PR #203, opened 2022-12-20
naively copied all examples over from https://github.com/es92/zkApp-e…

nicc:
…xamples


mitschabaude:
this can be closed, right?

PR #209, opened 2022-12-24
Add tutorial runner and test the first tutorial with it

rjdellecese:
To see this in action locally, you can use `act`—just [install it](https://github.com/nektos/act#installation) and then run `act --job=hello-world`.

You can also see the GitHub Actions run in my forked repo (https://github.com/rjdellecese/docs2/actions/workflows/test-tutorials.yml).


docs/zkapps/tutorials/01-hello-world.mdx
```diff
@@ -63,8 +63,19 @@ Now that you have the tooling installed, we can start building our application.
...
+<!--
+
+Configure `git` email and name, otherwise `zkapp-cli` project generation will return a non-zero exit code.
+
+```sh
+$ git config --global user.email "test@example.com"
+$ git config --global user.name "Test"
+```
+
+-->
```
rjdellecese, line 75: Using code blocks in HTML comments, it's trivial to interleave the code blocks that the reader sees with code blocks that get executed by the runner (and thus tested), but which are hidden from the rendered doc.

docs/zkapps/tutorials/01-hello-world.mdx
```diff
@@ -88,26 +99,26 @@ We will start by creating files for our project and deleting the default files t
...
 
 The `zk file <name>` command created both `src/Square.ts` and `src/Square.test.ts` for us, but we won't worry about writing tests in this tutorial. We'll use `main.ts` as a script to interact with the smart contract and observe how it works. In later tutorials, we will see how to interact with a smart contract from the browser, like a typical end user, but for now we'll just use our `main.ts` script to do this.
 
 Now, let's open `src/index.ts` in a text editor and change it to look like below. This file contains all exports we want to make available for consumption from outside our smart contract project, such as from a UI.
 
-```
-import { Square } from './Square.js';
-
-export { Square };
+```ts src/index.ts
```
rjdellecese, line 118: This syntax is valid per the CommonMark spec. Docusaurus expects CommonMark markdown, and thus syntax highlighting continues to work great in the docs with this syntax.

The string after the triple backticks but before the newline is called an "info string" in the spec. See https://spec.commonmark.org/0.30/#info-string for more… info 😅

rjdellecese:
I think this is ready for a look @mitschabaude

rjdellecese:
If we like it, I can continue working on the other tutorials. I can also have a look at how hard it might be to copy over the project resulting from this script's execution to an `examples` directory, a la #203.

mitschabaude:
Cool :D Will look at it tomorrow! Is this intended to run in CI already?

mitschabaude:
Ooh I just saw the CI run in your repo. Amazing! Do you know what's missing for it to run in our CI? GH Actions not set up? @jasongitmail?

scripts/tutorial-runner.ts
```diff
@@ -0,0 +1,174 @@
...
+        } else {
+          return codeLineWithNum.substring(codeLineWithNum.indexOf(' ') + 1);
+        }
+      };
+      const codeLinesWithNums = code.split('\n');
+      const startLineNum = getLineNum(codeLinesWithNums[0]);
+      const codeLines = codeLinesWithNums.map((codeLineWithNum, index) => {
+        if (startLineNum + index !== getLineNum(codeLineWithNum)) {
+          throw (
+            'Line numbers must be contiguous per code block.\n' +
```
mitschabaude, line 110: It's fine for now, but I think it could be nice to remove this restriction in the future. It offers the possibility to leave out stuff in the code samples, to be more concise but still provide valuable context, like
```
4   class MyContract extends SmartContract {
...
9     @method myMethod() {
10      // some code we care about
11    }
12  }
```

mitschabaude (approved):
This is fantastic work, I love this as a framework for creating tutorials that are checked!

rjdellecese:
Woohoo, I'm glad you like it! 😄

I'm not sure why it's the workflow isn't running in this repo; I'd have guessed that it would at least give you (as someone with commit permissions) the option of running it here. Maybe because it's a workflow that hasn't been run here before? 🤷‍♂️

scripts/tutorial-runner.ts
rjdellecese: Good point, and that snippet in particular is a great example. I was originally thinking that each range could just be a different code block, but I think that would be weird and possibly confusing in the snippet you provided. I can look into updating this as I set this up for other tutorials!

jasongitmail:
Will close #209

rjdellecese:
I can't see the Vercel failure—is it something I need to address?

MartinMinkov:
> I can't see the details of the Vercel failure—is it something I need to address?

There was a [fix](https://github.com/o1-labs/docs2/pull/226) introduced that I suspect your branch doesn't have. Just merge the latest changes in `main` and you should be good!

rjdellecese:
Hmm, we've run into an interesting situation here. When trying to merge the latest `main` into this branch, I had to resolve merge conflicts in the first tutorial due to https://github.com/o1-labs/docs2/pull/215. I did so, but now the test that this PR introduces is failing for this tutorial because #215 updates it for SnarkyJS 0.8.0, which is not released yet. So the test is working—it's telling us that for the current version of SnarkyJS (0.7.3), the tutorial is broken.

One way to proceed here would be to wait until SnarkJS 0.8.0 is released, at which point this test should start passing again. Does anyone know when that might happen?

In the future, it might be a better to wait until a new SnarkyJS version is released to merge in updates to the tutorials. Right now, until the new SnarkyJS release, anyone who goes to https://docs.minaprotocol.com/zkapps/tutorials/hello-world and tries to follow that tutorial will get stuck at [this step](https://docs.minaprotocol.com/zkapps/tutorials/hello-world#initializing-our-smart-contract) (you can see the failure in CI in my repo, [here](https://github.com/rjdellecese/docs2/actions/runs/3850846972/jobs/6561440528)).

mitschabaude:
I think you're right, @rjdellecese. The correct process would've been to wait with merging the tutorial updates until the new version was released. Luckily, in the future your work will force us to follow the correct process :D

mitschabaude:
We were planning to release 0.8.0 within the next week (hopefully.. it depends on a redeploy of the testnet which has many people involved)

rjdellecese:
> We were planning to release 0.8.0 within the next week (hopefully.. it depends on a redeploy of the testnet which has many people involved)

@mitschabaude sounds good! I'll wait until that happens to proceed with this work, then.

MartinMinkov (approved):
This is awesome! I tested it locally myself with `act`, and other than the currently out-of-sync tutorials, it works as expected!

Just a question for you, other tutorials interact with the zkapp-cli that requires interactive input from the user. I'm not sure if `shelljs.exec()` can handle these cases. Just as a heads up, it might be useful to look at some solutions there if you haven't already thought about it! I suspect `child_process` could work, but I'm not sure.

Looks great to me. Awesome work! :D

rjdellecese:
Thanks @MartinMinkov 😄 

> Just a question for you, other tutorials interact with the zkapp-cli that requires interactive input from the user. I'm not sure if shelljs.exec() can handle these cases. Just as a heads up, it might be useful to look at some solutions there if you haven't already thought about it! I suspect child_process could work, but I'm not sure.

Thanks for the heads-up! I think you're right about `child_process` (https://github.com/shelljs/shelljs/wiki/FAQ#running-interactive-programs-with-exec) but I'll do some more research.

docs/zkapps/tutorials/01-hello-world.mdx
```diff
@@ -63,8 +63,19 @@ Now that you have the tooling installed, we can start building our application.
...
+```sh
+$ git config --global user.email "test@example.com"
+$ git config --global user.name "Test"
+```
+
+-->
+
 ```sh
-zk project 01-hello-world
+$ zk project 01-hello-world
```
jasongitmail, line 78: @rjdellecese Is it possible to make this work without adding `$`? 

Removing this: 1.) cleans it up, particularly for newbies, b/c there's one less thing to visually parse, 2.) avoids getting copied--though we could update the JS copy code to strip it. 

Up to this point, the pattern we've followed for tutorials is to avoid using `$` unless the block contains _output_ from the command(s) as well. 

If it is required for you, then we can update the JS copy snippet and accept the added visual noise.

MartinMinkov:
Hey @rjdellecese, just another quick question! :D

How do you plan to address tutorials that have `...` that fill in content that was previously there? For example, take a look at this section:
https://docs.minaprotocol.com/zkapps/tutorials/common-types-and-functions#merkle-trees

I suspect this PR doesn't handle this case, but I could be mistaken. Wondering if you have any thoughts on how to adapt the current framework to address these points in the tutorials. Thank you!

docs/zkapps/tutorials/01-hello-world.mdx
rjdellecese: Yeah, I don't think that would be too hard. I personally like it for consistency's sake, though—in my experience inconsistency can be the most confusing thing for new learners, because when you lack context/prior knowledge, you generally start by looking for the simplest patterns you can find (such as "every shell command that I need to execute is preceded by a `$ `, rather than "sometimes `$ ` prepends a shell command that I need to execute, but sometimes it doesn't").

Another consideration is that if we updated the JS copy code to seek out lines prepended with `$ ` and then strip the `$ ` then it would also work with code blocks like those that currently intermix commands and output, which would be nice.

I'm still happy to change it if we'd prefer, just want to share a competing view! Also thanks for pointing out that this would break copying from certain shell code blocks, I had overlooked that.

rjdellecese:
> Hey @rjdellecese, just another quick question! :D
>
>  How do you plan to address tutorials that have ... that fill in content that was previously there? For example, take a look at this section:
https://docs.minaprotocol.com/zkapps/tutorials/common-types-and-functions#merkle-trees
>
> I suspect this PR doesn't handle this case, but I could be mistaken. Wondering if you have any thoughts on how to adapt the current framework to address these points in the tutorials. Thank you!

You're correct, this PR does not handle this case! And I have thought some about how to address that.

I'll write up more detailed thoughts tomorrow, but basically I think it's tricky but possible—the main trickiness (which applies to examples without `...` also) is that code insertions in the middle of a file need to know how to push existing code down. In other words, "inserting" (inserting code between lines) is different from "patching" (replacing code at given line numbers). Basically, I think we can come up with reasonable constraints that make it easy to do "inserting" properly for these tutorials. Worst-case, some diffing syntax (+/- signs prepending each line) could be an option, but I don't think that will be necessary.

MartinMinkov:
> > Hey @rjdellecese, just another quick question! :D
> > How do you plan to address tutorials that have ... that fill in content that was previously there? For example, take a look at this section:
> > https://docs.minaprotocol.com/zkapps/tutorials/common-types-and-functions#merkle-trees
> > I suspect this PR doesn't handle this case, but I could be mistaken. Wondering if you have any thoughts on how to adapt the current framework to address these points in the tutorials. Thank you!
> 
> You're correct, this PR does not handle this case! And I have thought some about how to address that.
> 
> I'll write up more detailed thoughts tomorrow, but basically I think it's tricky but possible—the main trickiness (which applies to examples without `...` also) is that code insertions in the middle of a file need to know how to push existing code down. In other words, "inserting" (inserting code between lines) is different from "patching" (replacing code at given line numbers). Basically, I think we can come up with reasonable constraints that make it easy to do "inserting" properly for these tutorials. Worst-case, some diffing syntax (+/- signs prepending each line) could be an option, but I don't think that will be necessary.

That sounds great! Thanks for your detailed reply. I just want to check, is this PR good to be merged? Just wanted to check if you had any other plans before I hit the button :P

rjdellecese:
@MartinMinkov I suppose it could be merged now, but if it does the GitHub Action will be failing until SnarkyJS `0.8.0` is released. I guess I have a slight preference for waiting until that happens so we can merge it in and then turn it on as a required status check, but ultimately I'm happy to defer to your preference!

I'd also like to make sure I address @jasongitmail's comment (https://github.com/o1-labs/docs2/pull/209#discussion_r1067309917), but we could decide to do so in a separate PR.

jasongitmail:
> I guess I have a slight preference for waiting until that happens so we can merge it in and then turn it on as a required status check
+1 waiting until 0.8.0

docs/zkapps/tutorials/01-hello-world.mdx
jasongitmail: > Another consideration is that if we updated the JS copy code to seek out lines prepended with $ and then strip the $ then it would also work with code blocks like those that currently intermix commands and output, which would be nice.
This is a great idea, and one we'll do. I'll create an issue because we can do this no matter which approach we take for single line commands. 

I'm still kicking around this idea. Your consistency argument is a fair point. An additional is that for any noobies who use their mouse to select/copy/paste, having only the command as selectable text is better for usability in such case.

rjdellecese: > An additional is that for any noobies who use their mouse to select/copy/paste, having only the command as selectable text is better for usability in such case.

Very good point as well.

jasongitmail: @rjdellecese Let's go with `$` for all, single and multi-line shell commands please. Either way will work, but just wanting to make a decision so this can move forward. We'll need to queue this PR until a corresponding PR is ready to update all tutorials to add the `$` to the code.
CC @ymekuria for that since he's working on the JS-snippet related code.

rjdellecese:
I see that SnarkyJS `0.8.0` has been released! 🎉 I'll get this updated tomorrow and hopefully we can get it merged soon thereafter.

docs/zkapps/tutorials/01-hello-world.mdx
jasongitmail: Update: All tutorials [have been updated](https://github.com/o1-labs/docs2/pull/247) to add `$` and the JS code copy snippet has been updated to strip it and any console output as well.

rjdellecese:
Looks like it's working! https://github.com/rjdellecese/docs2/actions/runs/3962889834/jobs/6790105449

rjdellecese:
Bump @jasongitmail @mitschabaude @MartinMinkov—I think this is ready to be merged.

nicc:
@mitschabaude Need to review update and merge

jasongitmail:
Merged! Thanks @rjdellecese!

PR #215, opened 2023-01-04
Update zkApps Example 01-04 and Tutorial 01

nicc:
...to align with SnarkyJS release 0.8.0


examples/zkapps/02-private-inputs-and-hash-functions/src/IncrementSecret.ts
```diff
@@ -4,22 +4,14 @@ import {
...
   @state(Field) x = State<Field>();
 
-  deploy(args: DeployArgs) {
-    super.deploy(args);
-    this.setPermissions({
-      ...Permissions.default(),
-      editState: Permissions.proofOrSignature(),
-    });
+  init() {
+    super.init();
```
mitschabaude, line 14: since `init()` doesn't do anything else than call the base `init()`, you can remove it

examples/zkapps/04-zkapp-browser-ui/contracts/src/Add.test.ts
```diff
@@ -8,6 +8,7 @@ import {
   PublicKey,
   AccountUpdate,
 } from 'snarkyjs';
+import { LocalBlockchain } from 'snarkyjs/dist/node/lib/mina';
```
mitschabaude, line 11: this should be imported from `'snarkyjs'`

examples/zkapps/04-zkapp-browser-ui/contracts/src/Add.test.ts
```diff
@@ -19,31 +20,36 @@ import {
 function createLocalBlockchain() {
   const Local = Mina.LocalBlockchain();
```
mitschabaude, line 21: nit: since this is a set of unit tests, I'd probably turn off proving with `proofsEnabled: false` here

examples/zkapps/04-zkapp-browser-ui/contracts/src/Add.test.ts
```diff
@@ -57,19 +63,20 @@ describe('Add', () => {
...
     const num = zkAppInstance.num.get();
-    expect(num).toEqual(Field.one);
+    expect(num).toEqual(Field(1));
   });
 
   it('correctly updates the num state on the `Add` smart contract', async () => {
     const zkAppInstance = new Add(zkAppAddress);
-    await localDeploy(zkAppInstance, zkAppPrivateKey, deployerAccount);
-    const txn = await Mina.transaction(deployerAccount, () => {
+    await Add.compile();
```
mitschabaude, line 73: if we turned off proving as suggested above, we can also remove `compile()` here (to make the tests run faster)

mitschabaude (approved):
excellent, approving in advance with a few comments!

docs/zkapps/tutorials/01-hello-world.mdx
```diff
@@ -187,29 +187,27 @@ Now, we will write the smart contract. Write the following in your file:
...
 
-In this case, we will allow our zkApp on-chain state to be altered by transactions containing account updates with _either_ proof or signature authorization. Proof authorization allows end users to alter the zkApp account state if their transaction was generated from your smart contract code. Signature authorization allows whomever deployed the smart contract to edit it.
+Since we're extending `SmartContract`, which has its own initialization to perform, we also need to call `super.init()` to invoke this function on the base class.
 
-In deploy, we also initialize our on-chain state, `num`, to a value of `3`.
+Then we initialize our on-chain state, `num`, to a value of `3`.
+
+We can also optionally [specify permissions](/zkapps/snarkyjs-reference/classes/SmartContract#setpermissions) here. The default permissions will allow our zkApp on-chain state to be altered by transactions containing account updates with _either_ proof or signature authorization. Proof authorization allows end users to alter the zkApp account state if their transaction was generated from your smart contract code. Signature authorization allows whomever deployed the smart contract to edit it.
```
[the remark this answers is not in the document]
nicc: Thanks! That sounds good.

examples/zkapps/02-private-inputs-and-hash-functions/src/IncrementSecret.ts
nicc: I was thinking it's worth keeping in just to illustrate initialisation. I find too much magic a little unhellpful in tutorials. But we do have Square.ts in Example 01 that shows a more realistic initialisation so maybe it's better to be concise here. So I'll remove it, thanks 😄

examples/zkapps/04-zkapp-browser-ui/contracts/src/Add.test.ts
nicc: Ah, thanks. Looks like it can be removed completely!

examples/zkapps/04-zkapp-browser-ui/contracts/src/Add.test.ts
nicc: Yeah I was thinking about this. I very muich agree in the context of a unit test but this is also an example. I wonder if too much test isolation comes at the expense of illustrating the system. wdyt?

mitschabaude: I think some of the other tutorials, where code examples aren't Jest tests, may be better places to show real compile / prove

PR #216, opened 2023-01-04
Update zkApps Example 05

nicc:
...to align with SnarkyJS release 0.8.0


examples/zkapps/05-common-types-and-functions/src/BasicMerkleTreeContract.ts
```diff
@@ -4,26 +4,16 @@ import {
...
   @state(Field) treeRoot = State<Field>();
 
-  deploy(args: DeployArgs) {
-    super.deploy(args);
-    this.setPermissions({
-      ...Permissions.default(),
-      editState: Permissions.proofOrSignature(),
-    });
+  init() {
+    super.init();
```
mitschabaude, line 16: remove empty `init()`

examples/zkapps/05-common-types-and-functions/src/LedgerContract.ts
```diff
@@ -18,12 +16,8 @@ class MerkleWitness20 extends MerkleWitness(20) {}
...
   @state(Field) ledgerRoot = State<Field>();
 
-  deploy(args: DeployArgs) {
-    super.deploy(args);
-    this.setPermissions({
-      ...Permissions.default(),
-      editState: Permissions.proofOrSignature(),
-    });
+  init() {
+    super.init();
```
mitschabaude, line 20: remove empty `init()`

examples/zkapps/05-common-types-and-functions/src/main.ts
```diff
@@ -26,301 +25,295 @@ import {
...
-  const num1EqualsNum2: Bool = num1.toUInt64().equals(num2);
-  
-  console.log("// --------------------------------------");
-  console.log(`num1 == num2: ${num1EqualsNum2.toString()}`);
-  console.log(`Fields in num1: ${num1.toFields().length}`);
+const num1EqualsNum2: Bool = num1.toUInt64().equals(num2);
 
-  // --------------------------------------
+console.log("// --------------------------------------");
+console.log(`num1 == num2: ${num1EqualsNum2.toString()}`);
```
mitschabaude, line 38: nit with the entire tutorial: it uses
```ts
`some text: ${something.toString()}`
```
even though putting stuff inside string templates `toString`s them already, so we could just use 
```ts
`some text: ${something}`
```

but you don't have to change that now!

examples/zkapps/05-common-types-and-functions/src/main.ts
```diff
@@ -26,301 +25,295 @@ import {
...
 
-  class Point extends CircuitValue {
-    @prop x: Field;
-    @prop y: Field;
+// --------------------------------------
 
-    static fromField(x: Field, y: Field) {
-      return new Point(x, y);
-    }
+class Point extends CircuitValue {
```
mitschabaude, line 93: it would be nice if this used `Struct` instead of (deprecated) `CircuitValue`, but we can also save that for later

mitschabaude (approved):
Looks good to me, just a few comments!

examples/zkapps/05-common-types-and-functions/src/main.ts
nicc: Thanks! I think a little elegance goes a long way in example code so I'm in favour of making this change. Not all of them can be coerced automatically, however (e.g. bools). In those cases I've moved the `.toString()` calls to the const declaration so we've still managed to strip them from string templates as suggested.

examples/zkapps/05-common-types-and-functions/src/main.ts
nicc: Agreed! I'd need a quick intro or referece to Struct but yes let's do that another time

PR #250, opened 2023-01-22
Update 02-private-inputs-hash-functions.mdx

deadblackclover:


LucidSamuel (commented):
// Comment: Regarding line 39 update - In the provided code, the "prove()" function is not necessary.
In certain cases, the `prove()` function might be used to generate a proof for a specific transaction or to interact with the Mina network. However, in the provided code, I think there is no explicit usage or reference to the prove() function.

deadblackclover:
@LucidSamuel This example does not run without using `await deployTxn.prove();`

Error when executing `node build/src/main.js`:

```
/02-private-inputs-and-hash-functions/node_modules/snarkyjs/dist/node/bindings/compiled/_node_bindings/snarky_js_node.bc.cjs:7467
         throw err;
         ^

Error: "Assert_failure src/lib/transaction_logic/zkapp_command_logic.ml:1286:17"
    at Object.sendTransaction (snarkyjs/dist/node/lib/mina.js:249:24)
    at async sendTransaction (snarkyjs/dist/node/lib/mina.js:745:12)
    at async Object.send (snarkyjs/dist/node/lib/mina.js:176:24)
    at async file:///02-private-inputs-and-hash-functions/build/src/main.js:22:1

```

barriebyron:
@ymekuria I see that the example file already has this code on line 38
https://github.com/o1-labs/docs2/blob/main/examples/zkapps/02-private-inputs-and-hash-functions/src/main.ts#L38

this PR adds 
```
await deployTxn.prove();
```

Should we also add line 39, so the doc shows both "await" 

```
await deployTxn.prove();
await deployTxn.sign([deployerKey, zkAppPrivateKey]).send();
```

ymekuria:
> Should we also add line 39, so the doc shows both "await"
> 
> ```
> await deployTxn.prove();
> await deployTxn.sign([deployerKey, zkAppPrivateKey]).send();
> ```

I don't think it is crucial to show because it is in the example code, but it wouldn't hurt to add it.

PR #260, opened 2023-02-03
add copy changes reflecting new wallet policy for MF delegation program

iregina:


docs/node-operators/foundation-delegation-program.mdx
```diff
@@ -135,18 +135,30 @@ See the following uptime leaderboards for the latest uptime performance scores:
...
+:::
 
 #### Payout Frequency
 
-Rewards must be distributed at least once for a given epoch (but you can also send them more frequently).
+Rewards must be distributed at least once for a given epoch.  You can send rewards more frequently, but the preferred method is to send one payment in the amount of your obligation.  This is the easiest method to avoid confusion in tracking payments, and will reduce the likelihood you will be incorrectly flagged as delinquent.
 
 All the rewards for epoch N must be delivered (ie. accepted in a block, not just sent) no later than slot number 3,500 of the next epoch. This gives you about a week to sort out these payments.
 
+:::tip
```
jasongitmail, line 158: @iregina Using 'Tip' takes it out of the flow of the writing, so probably doesn't make sense to use "Finally, ..." because it doesn't read as an item following a list above. 

Maybe this tip should just say something like: "Include a memo with your transaction that indicates which epoch number your payment corresponds to."

jasongitmail (changes requested):

docs/node-operators/foundation-delegation-program.mdx
iregina: Makes sense! Just updated.

jasongitmail (approved):

PR #264, opened 2023-02-09
Document how to capture assertion exceptions and show in a zkApp UI

### LINKED ISSUE #232: Document how to capture assertion exceptions and show in a zkApp UI
We need to document a suggested pattern, for how to capture and expose assertion failures to show helpful error messages in the UI to end users. Should be documented on the the [How to Write a zkApp UI](https://docs.minaprotocol.com/zkapps/how-to-write-a-zkapp-ui) page.

E.g. From Gregor: 
```js
try {
  callMyMethod(arg);
} catch (e) {
  switch (e.message) {
    case INSUFFICIENT_BALANCE:
      // show a modal telling the user they have insufficient balance for this
      break;
    // etc
  }
}
```

ymekuria:
Closes #232 

This PR adds a section at the end of the section [How to Write a zkApp UI](https://docs.minaprotocol.com/zkapps/how-to-write-a-zkapp-ui) on how to capture assertion errors and display helpful messages in the UI.

We can link to [this](https://github.com/o1-labs/docs2/issues/263) page documenting all the assertion exceptions and ways to mitigate them when it is created.

### DIFF
--- a/docs/zkapps/how-to-write-a-zkapp-ui.mdx
+++ b/docs/zkapps/how-to-write-a-zkapp-ui.mdx
@@ -231,6 +231,7 @@ href="https://docs.aurowallet.com/general/reference/api-reference/mina-provider-
 If an assertion error occurs while a user is interacting with your zkApp, you may want capture the assertion error and display a helpful message for the user in your UI.
 
 1. Capture errors that occur when a user interacts with a method on your smart contract in a try catch statement.
+2. Filter by the error message that was created.
 
 ```ts
 try {
--- a/docs/zkapps/how-to-write-a-zkapp-ui.mdx
+++ b/docs/zkapps/how-to-write-a-zkapp-ui.mdx
@@ -227,3 +227,5 @@ href="https://docs.aurowallet.com/general/reference/api-reference/mina-provider-
 :::
 
 ## Displaying assertion exceptions in your UI
+
+If an assertion error occurs while a user is interacting with your zkApp, you may want capture the assertion error and display a helpful message for the user in your UI.
--- a/docs/zkapps/how-to-write-a-zkapp-ui.mdx
+++ b/docs/zkapps/how-to-write-a-zkapp-ui.mdx
@@ -240,7 +240,8 @@ try {
 } catch (err) {
   let uiErrorMessage;
   switch (err.message) {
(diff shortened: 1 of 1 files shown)


jasongitmail (approved):
Left a question, but approving.

Should we mention or show an example assertion that uses a custom error message of `INSUFFICIENT_BALANCE` so they understand how the pieces fit together? (I mean literally the assertion itself, and no other lines of surrounding code.)

docs/zkapps/how-to-write-a-zkapp-ui.mdx
```diff
@@ -226,3 +225,27 @@ href="https://docs.aurowallet.com/general/reference/api-reference/mina-provider-
...
+2. Filter by the error message that was created.
+3. Display a helpful error message for the user in your UI based on what assertion exception was captured. This could be a modal or a message in your UI.
+
+```ts
+try {
+  YourSmartContract.yourMethod();
+} catch (err) {
+  let uiErrorMessage;
+  switch (err.message) {
+    case 'Overflow': // The assertion error message thrown by invoking YourSmartContract.yourMethod when there is an insufficent balance.
```
jasongitmail, line 243: Wasn't your previous case text clearer?

ymekuria: It was clearer but `INSUFFICIENT_BALANCE` is not an actual assertion error that is thrown.

jasongitmail: I assumed it was a custom error message, which is why I suggested that we include a line showing people an assertion with this custom error message. 

IMO, Overflow is not clear and this highlights it needs to be changed elsewhere; seems better to not use it here.

ymekuria: I struggled to find a common assertion error a zkApp user might encounter that is relatively clear to show as an example. This highlights the deeper problem we have with error messages. This ambiguity can be mitigated to a point when we have a document mapping the errors to their causes that developers can reference.

jasongitmail: Then we really need to prioritize that! I'll add an issue to SnarkyJS repo

Meanwhile, can we use a custom error message here please? So it's clear to people.

jasongitmail (approved):

PR #269, opened 2023-02-16
Tweak assertion exception text on UI page for clarity

jasongitmail:
Edited for precision and clarity.


iregina (approved):
Looks good! this does sound more readable

PR #307, opened 2023-03-14
Tutorial 7: setPermissions deprecated, replaced with account.permissions.set()

wsdt:


barriebyron:
@wsdt  thanks for the PR and apologies for the looooonnnnng delay. We've reworked the tutorial so this update is no longer required. Keep the feedback and contributions coming

PR #308, opened 2023-03-14
.Gte deprecated: Replaced with greaterThan().assertTrue()

wsdt:


jasongitmail:
Thanks!

@jackryanservia Let's schedule in to update this tutorial for the latest SnarkyJS version

PR #310, opened 2023-03-15
Noticed an extra bracket in this code snippet

iregina:
I think there's an extra code snippet here. Can you confirm?


docs/zkapps/tutorials/02-private-inputs-hash-functions.mdx
iregina, line 1: Never mind? I understand the second bracket is referring to the IncrementSecret class... but when I was following along I didn't get i!

barriebyron:
@iregina  I seem to recall seeing a conversation about this somewhere, maybe I dreamed it?
can you ask in Slack and tag the @product-guardian in Slack this one should be easy to close, thanks for keeping an eye on our braces

iregina:
I created a ticket about a "missing bracket" in this tutorial. It turns out the bracket isn't missing, it's just.... a continuation / part of a code snippet. However, since I got confused by this, I was wondering if you think we should remove. Martin decided to close it for now!

PR #345, opened 2023-04-05
feat: Update actionHash to actionState to reflect changes in SnarkyJS

MartinMinkov:


docs/zkapps/advanced-snarkyjs/actions-and-reducer.mdx
```diff
@@ -11,7 +11,7 @@ Please note that zkApp programmability is not yet available on Mina Mainnet, but
 
 # Actions & Reducer
 
-Like events, **actions** are _public_ arbitrary information passed along with a zkApp transaction. However, actions give you an additional power: you can process previous actions in a smart contract! Under the hood, this is possible because we store a commitment to the history of dispatched actions on every account -- the **actionsHash**. It allows us to prove that the actions you process are, in fact, the actions that were dispatched to the same smart contract.
+Like events, **actions** are _public_ arbitrary information passed along with a zkApp transaction. However, actions give you an additional power: you can process previous actions in a smart contract! Under the hood, this is possible because we store a commitment to the history of dispatched actions on every account -- the **actionState**. It allows us to prove that the actions you process are, in fact, the actions that were dispatched to the same smart contract.
```
barriebyron, line 14:
```suggestion
Like events, **actions** are _public_ arbitrary information that are passed along with a zkApp transaction. However, actions give you additional power: you can process previous actions in a smart contract! Under the hood, this is possible because a commitment is stored to the history of dispatched actions on every account -- the **actionState**. It allows you to prove that the actions you process are, in fact, the actions that were dispatched to the same smart contract.
```

docs/zkapps/advanced-snarkyjs/actions-and-reducer.mdx
```diff
@@ -11,7 +11,7 @@ Please note that zkApp programmability is not yet available on Mina Mainnet, but
...
 
 Using actions and a "lagging state" pattern, you can write zkApps that can _process concurrent state updates by multiple users_. Besides that, we imagine all kinds of use cases where actions act as a built-in, "append-only" off-chain storage layer.
```
barriebyron, line 16:
```suggestion
Using actions and a "lagging state" pattern, you can write zkApps that can _process concurrent state updates by multiple users_. With this capability, you can imagine all kinds of use cases where actions act as a built-in, "append-only" off-chain storage layer.
```

docs/zkapps/advanced-snarkyjs/actions-and-reducer.mdx
```diff
@@ -54,21 +54,21 @@ let stateType = Bool;
...
+What we called `acc` above is now called `state`; we also have to pass in the state's type as a parameter. In addition, we have to pass in an `actionState` which refers to one particular point in the actions history. Like `Array.reduce`, `Reducer.reduce` takes a callback which has the signature `(state: S, action: A) => S`, where `S` is the `stateType` and `A` is the `actionType`. It returns the result of applying all the actions, in order, to the initial `state`. In this example, the returned `state` is `Bool(true)` because one of the actions in the list is `Field(1000)`. Reduce also returns the new actionState -- so you can store it for using it when you reduce the next batch of actions. One last difference to JS reduce is that this takes a _list of lists_ of actions instead of a flat list. Each of the sublists are the actions that were dispatched in one account update (e.g., while running one smart contract method).
```
barriebyron, line 71:
```suggestion
The `acc` shown earlier is now `state`; you also have to pass in the state's type as a parameter. In addition, you have to pass in an `actionState` which refers to one particular point in the action's history. Like `Array.reduce`, `Reducer.reduce` takes a callback that has the signature `(state: S, action: A) => S`, where `S` is the `stateType` and `A` is the `actionType`. It returns the result of applying all the actions, in order, to the initial `state`. In this example, the returned `state` is `Bool(true)` because one of the actions in the list is `Field(1000)`. Reduce also returns the new actionState -- so you can store it to use when you reduce the next batch of actions. One last difference to JS reduce is that this takes a _list of lists_ of actions instead of a flat list. Each of the sublists are the actions that were dispatched in one account update (for example, while running one smart contract method).
```

barriebyron: @MartinMinkov  let's use [present tense](https://developers.google.com/style/tense) and [timelessness](https://developers.google.com/style/timeless-documentation)

I want to learn more about how we can describe the feature without bounding with "now"

docs/zkapps/advanced-snarkyjs/actions-and-reducer.mdx
```diff
@@ -54,21 +54,21 @@ let stateType = Bool;
...
 
 An astute reader may have noticed that this is eerily similar to a standard "Elm architecture" -- this is because we have an instance of a scan over an implicit infinite stream of actions (though here that are aggregated in chunks). This is very similar to the problem that came up when processing transactions within the [Mina Protocol L1 with Snark Workers](https://minaprotocol.com/blog/fast-accumulation-on-streams). This may sound scary, but it should be familiar to web developers through its instantiation via the Redux library or more recently via the `useReducer` hook in React!
```
barriebyron, line 73:
```suggestion
As an astute reader, you might have noticed that this use of `state` is eerily similar to a standard "Elm architecture" -- this similarity is because an instance of a scan over an implicit infinite stream of actions (though here that are aggregated in chunks). This is very similar to the problem that came up when processing transactions within the [Mina Protocol L1 with Snark Workers](https://minaprotocol.com/blog/fast-accumulation-on-streams). This may sound scary, but it is familiar to web developers through its instantiation by using the Redux library or more recently by using the `useReducer` hook in React!
```

barriebyron (commented):
ask me any questions, as always you do a thorough job of explaining the updates and impacts (I see that you are already thinking ahead to the upcoming What's New in SnarkyJS blog post and provided a succinct PR title and description!)

PR #350, opened 2023-04-11
Improve permissions docs

mitschabaude:
This enhances the permissions docs. Originally the plan was to add additional docs for using signatures in smart contract methods which build on the permissions example, but that plan was deprioritized


barriebyron (commented):
I know you are in a draft mode... an early edit and my suggestions are helping me test out the new [docs style guide](https://github.com/o1-labs/docs2/wiki/Docs-Style-Guide)

barriebyron (approved):
good idea to merge as it is, we can iterate in the future, as always. Thanks @mitschabaude !

PR #355, opened 2023-04-12
Time-Locked Accounts doc for change to convert times and dates into global slots

barriebyron:
Updates for https://github.com/o1-labs/snarkyjs/issues/821 

This PR includes [Time-Locked Accounts](https://docs.minaprotocol.com/zkapps/advanced-snarkyjs/time-locked-accounts) doc updates for the interface to convert times and dates into global slots

- change the docs so that the cliff time is absolute
- include the updated image in https://github.com/o1-labs/docs2/issues/354

### DIFF
--- a/docs/zkapps/advanced-snarkyjs/time-locked-accounts.mdx
+++ b/docs/zkapps/advanced-snarkyjs/time-locked-accounts.mdx
@@ -11,9 +11,11 @@ Please note that zkApp programmability is not yet available on Mina Mainnet, but
 
 # Time-Locked Accounts
 
-Time-locking allows you to pay someone in MINA or custom other tokens subject to a vesting schedule. This means that the tokens are initially locked, and only become available for withdrawal after a certain time, or gradually according to a certain schedule.
+Time-locking allows you to pay someone in MINA or custom other tokens subject to a vesting schedule. This means that the tokens are initially locked and become available for withdrawal only after a certain time or are available gradually according to a specific vesting schedule.
 
-The zkApp feature that enables time-locking is the `timing` field, which is present on every account. It look like this:
+The zkApp feature that enables time-locking is the `timing` field that is present on every account. 
+
+The `timing` field looks like this:
 
 ```ts
 type Account = {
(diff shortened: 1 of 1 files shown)


docs/zkapps/advanced-snarkyjs/time-locked-accounts.mdx
```diff
@@ -41,21 +43,21 @@ This graph shows how each of the timing properties affect the vesting schedule:
...
 
-As shown, the maximum amount of unlocked tokens is defined by the `initialMinimumBalance`. It is called "initialMinimumBalance" because, even though the tokens show up in the balance, they can't be withdrawn — in other words, the account has a a non-zero _minimum balance_. Initially, that minimum balance is equal to the amount of tokens locked (i.e. the initial minimum balance). Over time, the minimum balance decreases until it hits zero, which is the condition that makes `isTimed` false again.
+As shown, the maximum amount of unlocked tokens is defined by the `initialMinimumBalance`. The property is called "initialMinimumBalance" because, even though the tokens show up in the balance, they can't be withdrawn. The account has a a non-zero _minimum balance_. Initially, that minimum balance is equal to the amount of tokens locked (for example, the initial minimum balance). Over time, the minimum balance decreases until it hits zero, which is the condition that makes `isTimed` false again.
```
mitschabaude, line 48:
```suggestion
As shown, the maximum amount of unlocked tokens is defined by the `initialMinimumBalance`. The property is called "initialMinimumBalance" because, even though the tokens show up in the balance, they can't be withdrawn. The account has a a non-zero _minimum balance_. Initially, that minimum balance is equal to the amount of tokens locked (i.e., the initial minimum balance). Over time, the minimum balance decreases until it hits zero, which is the condition that makes `isTimed` false again.
```

Hey @barriebyron, **i.e.** (= id est) isn't meant in the sense of "for example", but in the sense of "that is" or "in other words". Not to be confused with "e.g.". Sometimes "i.e." can be replaced with "for example" but here it can't.


> Initially, that minimum balance is equal to the amount of tokens locked (i.e. the initial minimum balance).

This sentence was there to explain why we use the name `initialMinimumBalance` to refer to the total amount of tokens locked.

barriebyron:
closing this oldie that is now replaced by the new goodie https://github.com/o1-labs/docs2/pull/415

PR #368, opened 2023-04-24
remove warning message and revert back to old way to sending back rew…

iregina:
Purpose of this PR: Originally, the foundation DP had a set up where bps need to send rewards back to a specific address. However, now, they are reverting back to the original set up where BPs send back to the address that was delegated to them.

### DIFF
--- a/docs/node-operators/foundation-delegation-program.mdx
+++ b/docs/node-operators/foundation-delegation-program.mdx
@@ -136,18 +136,9 @@ See the following uptime leaderboards for the latest uptime performance scores:
 
 #### Payout Addresses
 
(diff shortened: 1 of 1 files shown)


barriebyron (approved):
@iregina  the text looks good, do we have supporting material that provides more context to the request for this change? 

Are there additional individuals that we need to tag here to provide insights and approvals? @jasongitmail and who else?

iregina:
> @iregina the text looks good, do we have supporting material that provides more context to the request for this change?
> 
> Are there additional individuals that we need to tag here to provide insights and approvals? @jasongitmail and who else?

The stakeholder for this is David Sedgwick and I just sent him the Vercel Preview link to see if the doc changes are approved.

PR #369, opened 2023-04-24
Fix newsletter footer form submission to properly submit email to MinaProtocol Newsletter

MartinMinkov:
# Description

Updates the footer to handle the proper form submission.

## Quick Question
When submitting to the newsletter on the `minaprotocol.com` website, the passed-in form data is encoded as the following:

image
The email is encoded as `input_1` for the React form submission as well.
https://github.com/o1-labs/docs2/pull/369/files#diff-f68519d7699db406dc369adc092d9efc0c065428f57c44b780f8378a5c19d163R32 . 

I am not sure if this will work as I cannot test locally since I run into CORS issues when submitting to the newsletter from localhost.

### DIFF
--- a/src/theme/Footer/index.tsx
+++ b/src/theme/Footer/index.tsx
@@ -24,19 +24,19 @@ function Footer(): JSX.Element | null {
     if (!email) {
       return;
     }
-    let response = await fetch(FormSubmitUrl, {
+    const response = await fetch(FormSubmitUrl, {
       method: 'POST',
       headers: {
         'Content-Type': 'application/x-www-form-urlencoded',
       },
-      body: `email=${email}`,
+      body: `input_1=${email}`, // Seems like this is the input field name for the email field on minaprotocol.com
     });
-
     if (!response.ok) {
       console.error('Failed to submit form');
       return;
     }
     setIsSubmitted(true);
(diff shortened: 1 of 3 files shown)


barriebyron:
hi @MartinMinkov  I am auto-subscribed as a reviewer, but do not have technical context to offer any insights.

However, I do like to spell `GitHub` correctly

src/theme/Footer/index.tsx
```diff
@@ -1,31 +1,42 @@
...
+      },
+      body: `email=${email}`,
+    });
+
+    if (!response.ok) {
+      console.error('Failed to submit form');
+      return;
+    }
     setIsSubmitted(true);
-    // console.log(event, email);
```
ymekuria, line 28: The `isSubmitted` flag can still be set to true here if the response does not return ok.

ymekuria (approved):
Nice Martin! Everything looks good assuming the submit network request works.

src/theme/Footer/index.tsx
MartinMinkov: Hmmm, I am not following. Doesn't the `if(!response.ok)` branch return early so it wouldn't trigger the state change?

ymekuria: Ha opps. I missed the return statement in your code. My mistake. This works great!

PR #375, opened 2023-04-26
Add `ignore` flag to docs tutorial runner and add better error message

MartinMinkov:
# Description
If we use a generic TypeScript code block with a ts annotation, the tutorial checker expects a file path to verify if the code is valid TypeScript. However, sometimes we only want to show generic TypeScript code blocks as examples of using an API. In this case, the tutorial runner will detect the TypeScript code block and produce an error because it expects a file path.

In this update, we add an ignore flag that can be used when we want to include generic TypeScript code blocks in tutorials without causing errors. We also improve the error messages to help identify which code block is causing the problem.


ymekuria (approved):
LGTM!

PR #395, opened 2023-05-13
Update 04-zkapp-ui-with-react.mdx

br0wnD3v:
Nextjs recommends the user to user App Router and might cause new comers to face difficulty during project scaffold `zk project [NAME] --ui next`


ymekuria:
@br0wnD3v thanks for bringing this up.  Try to update your cli version to 0.8.2. The app router feature is disabled in the current version and will be supported in the future. https://github.com/o1-labs/zkapp-cli/issues/400. We can update the zkApp cli version number that this tutorial is compatible with to avoid confusion.

PR #413, opened 2023-05-30
Fix readme link in contributing

Isaac-DeFrain:


barriebyron (approved):
thank you for the fix!

PR #423, opened 2023-06-12
Add embedded youtube links 

MartinMinkov:
# Description 
Instead of the old practice of using a manual thumbnail to specify a youtube link, we add the embedded element directly in our docs. This will improve UX and SEO.

## Steps to add an embedded video
If you want to add a youtube video in our docs, we can directly add the iframe HTML in our docs! :smile: To do so, we simply need to do the following:
1. Go to the youtube video you would link to link
2. Click _Share_
image
3. Click _Embed_
image
4. Click _Copy_
image
5. Paste the HTML in your clipboard directly in the docs!

### Things to note
You can resize the iframe by editing the HTML. For example, in the iframe element, there will be a height and width property. You can change these values to resize how big the iframe element is:

```html
<iframe
  width="900"
  height="550"
  src="https://www.youtube.com/embed/GvwYJDzzI-g"
 ...
></iframe>

```

### DIFF
--- a/docs/about-mina/consensus.mdx
+++ b/docs/about-mina/consensus.mdx
@@ -10,7 +10,13 @@ Consensus is the process by which the network determines which information is re
 
 Learn more about how Ouroboros Samasika works in this video:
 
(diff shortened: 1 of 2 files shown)


barriebyron:
@MartinMinkov This method is excellent upgrade to our video embed workflow! Thank you for the quick solution 
I see how you changed 
https://docs2-git-feat-fix-youtube-links-minadocs.vercel.app/about-mina/what-are-zero-knowledge-proofs

barriebyron (approved):
so good, so fast, so awesome!

PR #443, opened 2023-06-22
fix: replace ClientSDK with Mina Signer in client-sdk.mdx

barriebyron:
The ClientSDK is now deprecated and is superseded by MinaSigner https://github.com/o1-labs/snarkyjs/tree/main/src/mina-signer


barriebyron:
@MartinMinkov  do we need a redirect here? and if so, how do redirects work?

MartinMinkov:
> @MartinMinkov do we need a redirect here? and if so, how do redirects work?

Yes, good catch! You can add one by adding the following code in our `docusaurus.config.js` [file](https://github.com/o1-labs/docs2/blob/736d30e171bcaabe217a4121c1593b1d42d88f1d/docusaurus.config.js#L30C9-L30C9)

```js
{
  from: '/node-operators/client-sdk',
  to: '/node-operators/mina-signer',
},
```

MartinMinkov (approved):

PR #454, opened 2023-06-30
quick fix to clarify later version of a dependency

barriebyron:
The sentence wasn't clear about what versions might be later than


iregina (approved):
Looks great! Thanks for the clarity.

PR #458, opened 2023-07-03
fix incorrect variable name

duncand0nuts:
change old reference to variable num to x


barriebyron:
thanks for the update @duncand0nuts 
@Trivo25  can you verify this change and teach me?

Trivo25:
Looks good to me! Here's some context for you @barriebyron: When you check the source code right below that little piece of text, you can see it's referencing `x`, but the description above was talking about `num` - now it's fixed and less confusing!!

Trivo25 (approved):

PR #491, opened 2023-07-11
fix: glossary remove term without a definition

barriebyron:
Quick fix to add a link and remove an undefined term. 
More work to follow!

### DIFF
--- a/docs/glossary.mdx
+++ b/docs/glossary.mdx
@@ -1,6 +1,6 @@
 ---
 title: Glossary
-description: A zero knowledge list of terms with a focus on the Mina ecosystem. 
+description: A list of terms relating to zero knowledge and the Mina ecosystem. 
 keywords:
   - zk
   - zero knowledge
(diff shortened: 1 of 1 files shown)


docs/glossary.mdx
```diff
@@ -1,6 +1,6 @@
 ---
 title: Glossary
-description: zkApps (zero knowledge apps) are Mina Protocol smart contracts powered by zero knowledge proofs, specifically using zk-SNARKs. Use this quickstart guide to deploy zkApps and resources for learning TypeScript.
+description: A zero knowledge list of terms with a focus on the Mina ecosystem.
```
nicc, line 3: I think 'zero knowledge list' is fun but not so accurate and perhaps a bit of a stretch. What do you think of this suggestion?

```suggestion
description: A list of terms relating to zero knowledge and the Mina ecosystem. 
```

barriebyron: way better, it's always better together, thank you!

nicc (approved):

PR #494, opened 2023-07-12
Update README.md

LucidSamuel:
Think we should indicate the need to install docusaurus because ```npm run dev``` wouldn't work according to the code without doing so.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,6 +36,7 @@ After your local build environment is set, you can run a local build:
 ```sh
 git clone https://github.com/o1-labs/docs2.git
 npm install
+npm install docusaurus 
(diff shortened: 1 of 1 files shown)


barriebyron (approved):
thanks @LucidSamuel  for this update!

MartinMinkov:
For what it's worth, I don't run into this issue at all :/ The CLI tool is only used for scaffolding the project, it shouldn't affect starting the dev server

barriebyron:
well, if we don't need users to install docusaurus, then let's not have them do it

Revert PR https://github.com/o1-labs/docs2/pull/497

PR #508, opened 2023-07-16
Update generating-a-keypair.mdx

SanabriaRusso:
Path to my-wallet directory wrongly assumed to be under ~/. It was changed to $(pwd), as when mounting the directory to the container.


LucidSamuel:
@barriebyron The proposed change uses the $(pwd) variable, which expands to the current working directory. This ensures that the my-wallet directory will be properlychmod'd regardless of its location.

The original code assumed that the my-wallet directory would be located under the user's home directory (~/). However, this is not always the case. For example, if the directory is mounted to the container from a different location, the original code would not work.

barriebyron:
@LucidSamuel  if you think it's good to go, please approve and merge the PR

Thank you!

PR #532, opened 2023-07-21
429 update fee payer workflow in zkApp topics

barriebyron:
This PR updates the basic write a zkApp and deploy a zkApp with the new fee payer workflow

I want feedback on how much explainer content we need for the basic workflow, for this example, I guide the user to select a new account


barriebyron:
@ymekuria wdyt about the level of detail we give for the fee payer workflow? do we need a bigger, different explainer?
@LucidSamuel  wdyt?

ymekuria (approved):
Nice! I had one comment

PR #533, opened 2023-07-22
fix zkappWorker links and path

duncand0nuts:


ymekuria:
Thanks for adding this @duncand0nuts. There is another PR that updates all the links and the tutorial code here https://github.com/o1-labs/docs2/pull/526. I'm going to redeploy the code and this PR will be merged early next week.

barriebyron:
hi @duncand0nuts  thank you for fixing up the docs! Like @ymekuria  mentioned, we have the larger tutorial 4 updates happening in https://github.com/o1-labs/docs2/pull/526 

we'll look back here after that is merged and see how they fit together! thank you thank you

PR #541, opened 2023-07-26
Fix code snippet indentation in connecting-to-network pages

burnto:
Fix up a few places where the markup indentation was a little off. For example:

### Before

CleanShot 2023-07-26 at 15 23 46@2x

### After

CleanShot 2023-07-26 at 15 24 17@2x


barriebyron (approved):
thank you @burnto  for this docs love

PR #550, opened 2023-08-02
fix: How zkApps Work off-chain content

barriebyron:
how does off-chain storage work with zkApps? 
what do we want our devs to do for off-chain state?

This PR is a companion to the upcoming PR work for Tutorial 6: Off-Chain Storage 

Preview https://docs2-git-how-zkapp-fix-minadocs.vercel.app/zkapps/how-zkapps-work


bkase (approved):

barriebyron:
let's merge this quick fix and continue updates after a larger discussion about off-chain storage

PR #556, opened 2023-08-03
fix: move fetching events and actions to follow primary docs

barriebyron:
This PR fixes the wrong order of docs pages: Fetching events and actions was explained before events and actions
image


ymekuria (approved):
LGTM!

PR #566, opened 2023-08-17
Add tutorial 02 to CI

Trivo25:
The [Tutorial Runner](https://github.com/o1-labs/docs2/wiki/Tutorial-Runner) was implemented only for Tutorial 1; this PR adds it to Tutorial 2


.github/workflows/test-tutorials.yml
```diff
@@ -17,3 +17,17 @@ jobs:
...
           npx ts-node scripts/tutorial-runner.ts docs/zkapps/tutorials/01-hello-world.mdx
+
+  private-inputs-and-hash-functions:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v3
+        with:
+          # In case of version change, update README.md accordingly
+          node-version: 16
```
ymekuria, line 28: We probably should update the node version here and with the other tutorials in CI because node 16 has reached end of life.

ymekuria (approved):
Nice @Trivo25! This will be helpful to have. I just had a comment about upgrading the Node version here and for the other tutorials

.github/workflows/test-tutorials.yml
barriebyron: Yep, @shimkiv has landed https://github.com/o1-labs/zkapp-cli/pull/479/

barriebyron (approved):
LGTM I fixed conflicts in https://github.com/o1-labs/docs2/pull/566/commits/fe89c95bad235ec8c473ccc91317c6eaa4579f7f 

(due to this work using the previous name that we will no longer name)

.github/workflows/test-tutorials.yml
barriebyron: what is the right update here?

barriebyron:
@Trivo25  same comment as https://github.com/o1-labs/docs2/pull/577#issuecomment-1793052284

> I lost track of the status of this PR... were we waiting to merge it until the dependency on line numbers in the tutorial runner is removed? If we are waiting, wdyt we can move the PR to draft?

dkijania:
@Trivo25 are we still want to merge it since we have o1js docs separated now ?

dkijania:
Closing as obsolete, with thanks — this was right at the time.

This PR extends the `npx ts-node scripts/tutorial-runner.ts` job in `test-tutorials.yml`. That job no longer exists: `70915b36` ("ci: add tutorial tests", Oct 2024) replaced the runner with a matrix that builds and executes the checked-in projects under `examples/zkapps/`. The coverage this PR adds is already in place by other means — the tutorial it targets is in the run matrix today — and the accompanying `.mdx` changes reformat the page into runner syntax (line-numbered blocks, dedented code, hidden setup blocks) for a runner nothing invokes.

The goal behind it is carried forward in **#1249**, which takes it further: a project for every tutorial including the three that have none, every project run rather than build-only, tests in the nine packages that ship an empty jest harness, CI on pull requests, and a doc↔code drift check without the line-number coupling.

PR #575, opened 2023-08-31
fix: cleanup and delete unused image files

barriebyron:
This PR removed unused image files in the /docs2 repo

I searched the .mdx files that would call the removed images. 

I don't think the color Discord image files are used, since https://github.com/o1-labs/docs2/blob/main/src/theme/Footer/index.tsx used other image files

Docs still build with no complaints, so I think these files can simply be removed with this PR. Lots of old snap images, replaced with zkApp images


ymekuria (approved):
LGTM.

PR #579, opened 2023-09-06
make docs node v18 compatible 

### LINKED ISSUE #485: Upgrade NodeJS version from version 16
The site expects to be built with Node 16 and has issues using versions greater than that. The [end of life of Node 16 is approaching](https://nodejs.org/en/blog/announcements/nodejs16-eol), so we should have a plan to migrate to a higher version.

Trivo25:
this makes sure the project builds with node >=18 (LTS)

also updates outdated packages

closes #485


package.json
```diff
@@ -6,7 +6,7 @@
     "docusaurus": "docusaurus",
     "start": "docusaurus start",
     "dev": "ALGOLIA_APP_ID='.' ALGOLIA_SEARCH_API_KEY='.' docusaurus start",
-    "build": "docusaurus build",
+    "build": "NODE_OPTIONS=--openssl-legacy-provider && docusaurus build",
```
ymekuria, line 9: Is `NODE_OPTIONS=--openssl-legacy-provider` a workaround to handle deprecations or breaking changes in newer versions of Node.js related to OpenSSL?

ymekuria (approved):
Thanks for doing the @Trivo25. I just have a question.

package.json
Trivo25: Actually, after doing some more research: this should not be required. Need @MartinMinkov to check if Vercel uses the latest LTS v18 to build the project, if so, I can revert that workaround

README.md
```diff
@@ -17,14 +17,15 @@ During the documentation build process, each PR goes through Vercel CI to test t
 ## How to see your local doc updates during development
 
 To test the doc build and see local updates before you push changes to GitHub, be sure to build the docs locally.
+
 <!-- In case of version change, update .github/workflows/{test-tutorials.yml, snarkjs-api-reference.md} accordingly -->
 
-Local builds require Node.js, version 16 precisely.
+Local builds require Node.js, version 18 precisely.
```
barriebyron, line 23: @Trivo25 thanks for this PR.
Do docs build require node 18 precisely?

MartinMinkov (approved):

barriebyron (approved):
thank you for this PR!

PR #620, opened 2023-10-06
fix: remove unused term Kademlia, add more Mina terms

barriebyron:
This PR
- removes a term that is no longer relevant
- adds Mina terms that are relevant
- fixes a busted in-doc link and a syntax error


ymekuria (approved):
Nice additions!

PR #621, opened 2023-10-06
fix: move zkApp example code for Tutorial 9: Recursion to docs2 repo

barriebyron:
The example code for https://docs.minaprotocol.com/zkapps/tutorials/recursion was in Evan's private repo
This PR:
- copies the folder from https://github.com/es92/zkApp-examples/blob/main/09-recursion/src/vote.ts
- copies other files in the folder from our own tuto 1

Please verify that the files and required other content is accurate for o1js


ymekuria (approved):
The only thing to note is we have to update SnarkyJS to o1js and make sure the most current versions are used.

PR #626, opened 2023-10-06
Fixed using wrong mina-signer method

jackryanservia:
Was using wrong mina-signer method in Vercel oracle and getting invalid signatures.


barriebyron (approved):
LGTM... does this updated code show anywhere in the docs for this tuto? if no, you're good to go

If yes, let's update the doc too

ymekuria (approved):

PR #631, opened 2023-10-09
fix: delete duplicate Getting Started topic

barriebyron:
This PR deletes a duplicate Getting Started topic that is not present in https://github.com/o1-labs/docs2/blob/main/sidebars.js 

The canonical Getting Started content is published for Node Operators https://docs.minaprotocol.com/node-operators/getting-started


ymekuria (approved):
LGTM!

Issue #476, opened 2023-07-03
HF audit zkApp docs: Interacting With Mina /zkapps/o1js/interact-with-mina.md


barriebyron:
Audit and update all zkApp docs for technical accuracy prior to hard fork #459
Before the HF, verify technical content in

Interacting With Mina 
https://docs.minaprotocol.com/zkapps/snarkyjs/interact-with-mina 

462 lines
https://github.com/o1-labs/docs2/blob/main/docs/zkapps/snarkyjs/interact-with-mina.md

Consider enhancing value of this topic with purpose, links out to use cases, blog stories, and so on

PR #652, opened 2023-10-16
HF docs: verify Interacting with Mina 

barriebyron:
This PR verifies the content for HF for the Interacting with Mina doc
Closes [#476](https://github.com/o1-labs/docs2/issues/476) 

Please review in entirety. 

See my questions in the files changed view and lmk what changes we might need

Preview doc https://docs2-git-interact-mina-minadocs.vercel.app/zkapps/o1js/interact-with-mina

### DIFF
--- a/docs/zkapps/o1js/interact-with-mina.md
+++ b/docs/zkapps/o1js/interact-with-mina.md
@@ -43,7 +43,7 @@ For example, if you transfer MINA from one account to another, the balance on tw
 
 Account updates are a flexible and powerful data structure that can express all kinds of updates, events, and preconditions you use to develop smart contracts.
 
-### Transaction JSON object
+## Transaction structure
 
 A _transaction_ is a JSON object of the form: 
 
@@ -110,9 +110,7 @@ First of all, this is an array with two entries -- the two account updates that 
 
 The first entry is always the fee payer, whose public key was passed in as `sender`. For the `fee`, which you didn't specify, o1js filled in 0; the `authorization` was filled with a dummy signature. In a user-facing zkApp, you typically don't care about setting those values – instead, you create a transaction like this in the browser and pass it on to the user's wallet. The wallet replaces your fee payer with one that represents the user account, with the user's settings for the fee. It would also fill the `authorization` field with a signature created from the user's private key. See [connecting your zkApp with a user's wallet](../how-to-write-a-zkapp-ui#connecting-your-zkapp-with-a-users-wallet).
 
-The second account update has the `'MyContract.myMethod()'` label:
-
-- The update corresponds to the method call. An `@method` call always results in the creation of an account update – an update to the zkApp account itself. 
+The second account update has the `'MyContract.myMethod()'` label. The update corresponds to the method call. An `@method` call always results in the creation of an account update – an update to the zkApp account itself. 
 
 Other fields in this account update are:
 
@@ -264,8 +262,7 @@ In general, there are three kinds of authorizations that an account update can h
 
 - Proof authorization – used for zkApp accounts when you do an `@method` call. Proofs are verified against the on-chain verification key.
 - Signature authorization – used to update user accounts. Signatures are verified against the account's public key.
-- No authorization – used on updates which don't require authorization. For example, positive balance changes. 
-<!-- is it proof or signature only, why would we include "no authorization" here? is it only the state updates that require proof or signature?-->
+- No authorization – used on updates which don't require authorization. For example, positive balance changes.
 
 ## Account update tree structures
 
--- a/docs/zkapps/o1js/interact-with-mina.md
+++ b/docs/zkapps/o1js/interact-with-mina.md
@@ -143,11 +143,11 @@ await tx.prove(); // this might take a while
 
 This example code includes two new operations:
 
-- `MyContract.compile()` creates prover and verification keys from your smart contract. You must create the keys before you can create any proofs.
-
-  - The name `compile()` is a metaphor for what this function does: creating prover and verifier functions from your code. It doesn't refer to literal "compilation" of JS into a circuit representation. The circuit representation of your code is created by _executing_ it, not by compiling it. Also, the prover function still includes the execution of your JS code as one step.
+- `MyContract.compile()` creates prover and verification keys from your smart contract.[^1] You must create the keys before you can create any proofs.
 
 - `tx.prove()` goes through your transaction and creates proofs for all the account updates that came from method calls.
+
+[^1]:The name `compile()` is a metaphor for what this function does: creating prover and verifier functions from your code. It doesn't refer to literal "compilation" of JS into a circuit representation. The circuit representation of your code is created by _executing_ it, not by compiling it. Also, the prover function still includes the execution of your JS code as one step.
 
 Both of these heavy cryptographic operations can take between a few seconds and a few minutes, depending on the amount of logic you're proving and on how fast your machine is. If you print the transaction again with `tx.toPretty()`, it now has the proof as a Base64 string inside the `authorization` field:
 
@@ -257,7 +257,7 @@ Two quick observations:
 
 - The user update has `authorizationKind: 'None_given'`. That means the update is not authorized. This is possible because it doesn't include any changes that require authorization: It just receives MINA. You can send someone MINA without their permission.
 
-In general, there are three kinds of authorizations that an account update can have: a proof, a signature, or none. We'll learn about signatures in the next section. You can find more details in [our docs about permissions](TODO).
+In general, there are three kinds of authorizations that an account update can have: a proof, a signature, or none. You learn about signatures in the next section. You can find more details in [Permissions](/zkapps/o1js/permissions).
 
 
 - Proof authorization – used for zkApp accounts when you do an `@method` call. Proofs are verified against the on-chain verification key.
@@ -406,6 +406,7 @@ To create the sender private key in a script:
 
 ```ts
 const senderPrivateKey = PrivateKey.fromBase58('EKEQc95...');
+```
 
 In a real server-side deployment, you probably want to load keys from a file or environment variable, instead of hard-coding them in your source code.
 :::
--- a/docs/zkapps/o1js/interact-with-mina.md
+++ b/docs/zkapps/o1js/interact-with-mina.md
@@ -257,9 +257,7 @@ Two quick observations:
 
 - The user update has `authorizationKind: 'None_given'`. That means the update is not authorized. This is possible because it doesn't include any changes that require authorization: It just receives MINA. You can send someone MINA without their permission.
 
-In general, there are three kinds of authorizations that an account update can have: a proof, a signature, or none. You learn about signatures in the next section. You can find more details in [Permissions](/zkapps/o1js/permissions).
-
-
+In general, there are three kinds of authorizations that an account update can have: a proof, a signature, or none. You learn more about signatures in the next section. You can find more details in [Permissions](/zkapps/o1js/permissions).
 
 ## Account update tree structures
 
--- a/docs/zkapps/o1js/interact-with-mina.md
+++ b/docs/zkapps/o1js/interact-with-mina.md
@@ -260,9 +260,6 @@ Two quick observations:
 In general, there are three kinds of authorizations that an account update can have: a proof, a signature, or none. You learn about signatures in the next section. You can find more details in [Permissions](/zkapps/o1js/permissions).
 
 
-- Proof authorization – used for zkApp accounts when you do an `@method` call. Proofs are verified against the on-chain verification key.
-- Signature authorization – used to update user accounts. Signatures are verified against the account's public key.
-- No authorization – used on updates which don't require authorization. For example, positive balance changes.
 
 ## Account update tree structures
 
--- a/docs/zkapps/o1js/interact-with-mina.md
+++ b/docs/zkapps/o1js/interact-with-mina.md
@@ -1,7 +1,7 @@
 ---
 title: Interacting With Mina
 hide_title: true
-description: How to create zero knowledge proofs and how users call zkApp methods. An account update can have proof, signature or none authorizations.
+description: How to create zero knowledge proofs and how users can call zkApp methods. An account update can have proof or signature authorizations.
 keywords:
   - smart contracts
   - zkapps
@@ -27,15 +27,36 @@ zkApp programmability is not yet available on the Mina Mainnet. You can get star
 
 # Interacting With Mina
 
-## Transactions and account updates
+Now that you know about writing zkApp methods, it's time to learn how users can call these methods. 
 
-Now that you have an idea about writing zkApp methods, it's time to learn how users can call these methods. Recall that smart contracts execute off-chain. The result of such an off-chain execution is a _transaction_, which can be sent to the Mina network to apply the changes made by the smart contract. In this section, you learn what a transaction looks like, and how you can create one.
+## Transactions
 
-The fundamental data structure that Mina transactions are built from is called an _account update_. An account update always contains updates to one specific on-chain account. For example, if you transfer MINA from one account to another, the balance on two accounts is updated – the sender and the receiver. Therefore, sending MINA requires two account updates. Account updates are a flexible and powerful data structure that can express all kinds of updates, events and preconditions that you use for developing smart contracts.
+Recall that smart contracts execute off-chain. The result of an off-chain execution is a _transaction_ that can be sent to the Mina network to apply the changes made by the smart contract.
 
-A _transaction_ is a JSON object of the form `{ feePayer, accountUpdates: [...], memo }`. Here, the `feePayer` is a special account update of slightly simpler structure. In particular, it contains a `fee` field which has to be used to specify the transaction fee. The `accountUpdates` array, on the other hand, is a list of normal account updates, which make up the bulk of the transaction. Finally, `memo` is an encoded string which can be used to attach an arbitrary short message. Ignore it for now.
+In this section, you learn what a transaction looks like and how to create one.
 
-You create transactions in o1js by calling `Mina.transaction(...)`, which takes the sender (a public key) and a callback that contains your transaction's logic.
+## Account updates
+
+The fundamental data structure that Mina transactions are built from is called an _account update_. An account update always contains updates to one specific on-chain account. 
+
+For example, if you transfer MINA from one account to another, the balance on two accounts is updated – the sender and the receiver. Therefore, sending MINA requires two account updates. 
+
+Account updates are a flexible and powerful data structure that can express all kinds of updates, events, and preconditions you use to develop smart contracts.
+
+### Transaction JSON object
+
+A _transaction_ is a JSON object of the form: 
+
+```ts
+{ feePayer, accountUpdates: [...], memo }
+```
+
+- `feePayer` is a special account update with a slightly simpler structure. 
+  - In particular, it contains a `fee` field, which must be used to specify the transaction fee. 
+- `accountUpdates` is an array of a list of normal account updates that make up the bulk of the transaction. 
+- `memo` is an encoded string that can be used to attach an arbitrary short message.
+
+You create transactions in o1js by calling `Mina.transaction(...)`, which takes the sender (a public key) and a callback that contains your transaction logic:
 
 ```ts
 const sender = PublicKey.fromBase58('B62..'); // the user address
@@ -46,19 +67,23 @@ const tx = await Mina.transaction(sender, () => {
 });
 ```
 
-In this example, the transaction consists of calling a single `SmartContract` method, called `myMethod`. You can inspect the transaction yourself by printing it out as JSON:
+In this example, the transaction calls a single `SmartContract` method called `myMethod`. You can inspect the transaction by printing the JSON data to the console:
 
 ```ts
 console.log(tx.toJSON());
 ```
 
-If you try this, you see a massive JSON object with lots of fields, most of which are set to their default value. There's also a way to pretty-print transactions in a more human-readable, condensed format:
+This command outputs a massive JSON object with many fields, most of which are set to their default value. 
+
+### Transaction output
+
+Instead, it's better to output transactions in a more human-readable, condensed format to print only the fields with actual content:
 
 ```ts
 console.log(tx.toPretty());
 ```
 
-Depending on the logic of `myMethod()`, this could print something like the following:
+Depending on the logic of `myMethod()`, the output is something like:
 
 ```ts
 [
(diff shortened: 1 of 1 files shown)


docs/zkapps/o1js/interact-with-mina.md
```diff
@@ -1,7 +1,7 @@
 ---
 title: Interacting With Mina
 hide_title: true
-description: How to create zero knowledge proofs and how users call zkApp methods. An account update can have proof, signature or none authorizations.
+description: How to create zero knowledge proofs and how users can call zkApp methods. An account update can have proof or signature authorizations.
```
barriebyron, line 4: we have conflicting statements: proof, signature, AND none (or just proof or signature?) do we need an explainer?

docs/zkapps/o1js/interact-with-mina.md
```diff
@@ -81,25 +106,44 @@ Depending on the logic of `myMethod()`, this could print something like the foll
...
-- `authorization: undefined` – the proof that's needed on this update isn't there yet! You learn how to add it in a minute.
+Note that there are many more fields that account updates can have, but `tx.toPretty()` prints only the fields with actual content. Also, not all account updates must be present. For example, if a zkApp doesn't set a state, the `update` field might be missing. In that case, strictly speaking, it wouldn't always be an "update" in the sense that the account is modified. The term "account update" is used for simplicity.
 
-Note that there a many more fields that account updates can have, but `tx.toPretty()` only prints the ones that have actual content. Also, the ones above may be missing: For example, if our zkApp doesn't set any state, the `update` field might be missing. In that case, strictly speaking it wouldn't always be an "update" in the sense that the account is modified. The term "account update" is used for simplicity.
+As you might have noticed, account updates aren't created in a very explicit manner. Instead, o1js gives you an imperative API, with "commands" like `state.set()`, to create and modify account updates in a transaction.
```
barriebyron, line 140: what is "imperative API" mentioned here? just that (what is it called?) our API requires the "commands" and doesn't imply anything?

docs/zkapps/o1js/interact-with-mina.md
```diff
@@ -132,20 +177,33 @@ Both of these are heavy cryptographic operations and can take between a few seco
...
+
+- The verifier passes in the public input when verifying it.
+- The proof is valid only if it was created with _the same public input_. 
 
-You can think of the public input as data that is shared between the prover and verifier. The verifier passes in the public input when verifying it, and the proof is valid only if it was created with _the same public input_. This means that this proof attests to the validity of exactly this account update. If you change the account update before sending it to the Mina network, the proof is no longer not be valid. In other words: The only valid account updates for a zkApp account are the ones created according to the logic of your `SmartContract`. This is the core of why we can have smart contracts that execute on the client side.
+The proof attests to the validity of exactly this account update. If you change the account update before sending it to the Mina network, the proof is no longer valid. 
 
-## Payments and more on public inputs
+The only valid account updates for a zkApp account are the ones created according to the logic of your `SmartContract`. This core concept is why zkApp smart contracts execute on the client side.
```
barriebyron, line 200: is there a reason to use code literal here? Is the code always `SmartContract`? or we can just use plain text?

> your `SmartContract`

docs/zkapps/o1js/interact-with-mina.md
```diff
@@ -171,11 +233,11 @@ console.log(tx.toPretty());
 
 :::info
 
-MINA amounts, in all o1js APIs and elsewhere in the protocol, are always denominated in nanoMINA = `10^(-9)` MINA. This is why we set `const MINA = 1e9`.
+MINA amounts, in all o1js APIs and elsewhere in the protocol, are always denominated in nanoMINA = `10^(-9)` MINA (which is why we set `const MINA = 1e9`).
```
barriebyron, line 236: what does this mean? do we need to include it? and can we link somewhere to explain it?

> (which is why we set `const MINA = 1e9`).

docs/zkapps/o1js/interact-with-mina.md
```diff
@@ -171,11 +233,11 @@ console.log(tx.toPretty());
...
 
 :::
 
-What's interesting is that the transaction now has 3 account updates:
+The transaction now has these account updates:
```
barriebyron, line 240: I didn't see three updates, I saw two... so I changed the word here. Is two accurate?

docs/zkapps/o1js/interact-with-mina.md
```diff
@@ -420,43 +486,15 @@ await pendingTx.wait();
...
-  // ...
-});
-await tx.prove();
-tx.sign([senderKey]);
-
-await tx.send();
-
-// our accounts updates are applied locally!
-```
+Doing this means setting up a fresh, local ledger that is pre-filled with a couple of accounts with funds that you have access to. "Sending" a transaction here just means applying your account updates to that local Mina instance. This is helpful for testing, especially because account updates go through the same validation logic locally that they would on-chain.
```
barriebyron, line 496: do we want to provide more details here? what does it mean to set up a local ledger?

docs/zkapps/o1js/interact-with-mina.md
```diff
@@ -353,43 +406,53 @@ await tx.prove();
...
 
 ```ts
 const senderPrivateKey = PrivateKey.fromBase58('EKEQc95...');
-```
 
 In a real server-side deployment, you probably want to load keys from a file or environment variable, instead of hard-coding them in your source code.
-:::
 
-To summarize, there are three types of authorization that account updates can have, which are typically used in different circumstances:
+Recall that account updates can have three types of authorization:
```
barriebyron, line 430: we have more than three, right? Do we want to summarize here and link to https://docs.minaprotocol.com/zkapps/o1js/permissions#types-of-authorizations ?

docs/zkapps/o1js/interact-with-mina.md
```diff
@@ -281,35 +327,38 @@ class MyContract extends SmartContract {
...
 
 ```ts
 let senderUpdate = AccountUpdate.create(this.sender);
 ```
 
-`AccountUpdate` is the class in o1js that represents account udpates. `AccountUpdate.create()` not only instantiates this class, but also attaches the update to the current transaction, at the same level where `create` is called. If it is called inside a `@method`, the `AccountUpdate` is created as a child (public input) of the zkApp update.
+- `AccountUpdate` is the class in o1js that represents account udpates. 
+- `AccountUpdate.create()` instantiates this class and attaches the update to the current transaction at the same level where `create` is called. 
+
+If it is called inside an `@method`, the `AccountUpdate` is created as a child (public input) of the zkApp update.
```
mitschabaude, line 339: This belongs into the second bullet point above; "it" refers to `AccountUpdate.create()`

docs/zkapps/o1js/interact-with-mina.md
```diff
@@ -333,14 +382,18 @@ If we create a transaction for calling this method like before, it looks like th
...
 
-- `useFullCommitment: true`, which we won't explain here but has to do with replay protection if you're using signatures.
+- `useFullCommitment: true`, not explained here but has to do with replay protection when using signatures.
 - `authorizationKind: 'Signature'`
 
-Finally, `authorization: undefined` indicates that we didn't provide the signature yet.
+Finally, `authorization: undefined` indicates that the signature is not provided yet.
 
-In a user-facing zkApp, user signatures will typically be added by a wallet, not within o1js. In that case, the missing signature is fine. However, for testing and calling zkApps via node, you need to add the signatures yourself. The command to do that is `tx.sign([...privateKeys])`, called after `Mina.transaction` on the finished transaction. Here's an example:
+In a user-facing zkApp, user signatures are typically added by a wallet, not within o1js. In that case, the missing signature is expected. However, to use node for testing and calling zkApps, you must add the signatures with `tx.sign([...privateKeys])`, called after `Mina.transaction` on the finished transaction.
```
mitschabaude, line 394:
```suggestion
In a user-facing zkApp, user signatures are typically added by a wallet, not within o1js. In that case, the missing signature is expected. However, in tests or when calling zkApps from a Node.js script, you must add the signatures with `tx.sign([...privateKeys])`, called after `Mina.transaction` on the finished transaction. 
```

docs/zkapps/o1js/interact-with-mina.md
```diff
@@ -353,43 +406,53 @@ await tx.prove();
...
+:::tip
+o1js allows you to load and store private and public keys in Base58 format. 
+:::
 
-o1js allows you to load and store private and public keys in base58 format. Here's how the sender private key might be created in a script:
+To create the sender private key in a script:
 
 ```ts
 const senderPrivateKey = PrivateKey.fromBase58('EKEQc95...');
-```
```
mitschabaude, line 366: the backticks here are needed to end the code example, should be put back

docs/zkapps/o1js/interact-with-mina.md
```diff
@@ -353,43 +406,53 @@ await tx.prove();
...
+  
+  `.sign()` takes an array, so you could provide multiple private keys for signing. 
 
-Note that `.sign()` takes an array, so you could provide multiple private keys for signing. `.sign()` will go through the transaction and add signatures on all account updates which 1) need a signature and 2) whose public key matches one of the private keys that were provided. In the example above, two account updates are signed with `tx.sign()`: The fee payer and the depositor account update. Both have the `sender` public key on them, which matches `senderPrivateKey.toPublicKey()`.
+In this example, two account updates are signed with `tx.sign()`: The fee payer and the depositor account update. Both have the `sender` public key on them that matches `senderPrivateKey.toPublicKey()`.
 
-:::note
+:::tip
+o1js allows you to load and store private and public keys in Base58 format. 
+:::
```
mitschabaude, line 421: I think the code below should be part of this tip! Otherwise it is ripped from context

docs/zkapps/o1js/interact-with-mina.md
mitschabaude: No, the list in the Permissions docs is a slightly different notion of authorization types. What we discuss here are types of authorization that an account update can have, and the list of three below is exhaustive

docs/zkapps/o1js/interact-with-mina.md
```diff
@@ -407,7 +470,10 @@ tx.sign([senderPrivateKey]);
 await tx.send();
 ```
 
-The output of `tx.send()` can be used to wait for inclusion of this transaction in a block, and to get the transaction hash (which lets you look up the pending transaction on a block explorer):
+The output of `tx.send()` can be used to:
+
+- Wait for inclusion of this transaction in a block
+- To get the transaction hash, which lets you look up the pending transaction on a block explorer
```
mitschabaude, line 476: shouldn't this be

```suggestion
- Get the transaction hash, which lets you look up the pending transaction on a block explorer
```

docs/zkapps/o1js/interact-with-mina.md
```diff
@@ -420,43 +486,15 @@ await pendingTx.wait();
...
-tx.sign([senderKey]);
-
-await tx.send();
-
-// our accounts updates are applied locally!
-```
+Doing this means setting up a fresh, local ledger that is pre-filled with a couple of accounts with funds that you have access to. "Sending" a transaction here just means applying your account updates to that local Mina instance. This is helpful for testing, especially because account updates go through the same validation logic locally that they would on-chain.
 
-Note that with `proofsEnabled: false`, you still need to call `await tx.prove()`. The only change is that `.prove()` creates a dummy proof instead of a real one, and runs in 0 seconds. Also, `tx.send()` will skip the usual step of verifying your proof. -->
+Fun fact: The `LocalBlockchain` instance literally uses the same OCaml code for transaction validation and application that the Mina node uses; it's compiled to JS with [js_of_ocaml](https://github.com/ocsigen/js_of_ocaml).
```
mitschabaude, line 498: Don't we use footnotes? 😅 Like the one above, this feels more like footnote content to me (but having it in the main text works as well, in this case)

docs/zkapps/o1js/interact-with-mina.md
mitschabaude: The "ledger" is the list of accounts on the blockchain. Setting up a local ledger means creating a list of accounts on the developer's machine, in memory, instead of on an actual blockchain.

I don't think that explanation needs to be added, though.

docs/zkapps/o1js/interact-with-mina.md
```diff
@@ -81,25 +106,44 @@ Depending on the logic of `myMethod()`, this could print something like the foll
...
 
-From this output, there are several important things we can learn about transactions.
+This output includes several essential things to learn about transactions.
+
+The transaction output is an array with two entries: 
+
+  - The account updates that make up this transaction
+  - 'MyContract.myMethod()'` that corresponds to the method call you performed
+
+Account updates:
```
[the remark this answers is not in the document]
mitschabaude: see https://github.com/o1-labs/docs2/pull/652#discussion_r1361609448

docs/zkapps/o1js/interact-with-mina.md
barriebyron, line 4:
```suggestion
description: How to create zero knowledge proofs and how users can call zkApp methods. An account update can have proof, signature, or none authorizations.
```

docs/zkapps/o1js/interact-with-mina.md
```diff
@@ -46,19 +67,23 @@ const tx = await Mina.transaction(sender, () => {
...
 ```ts
 console.log(tx.toJSON());
 ```
 
-If you try this, you see a massive JSON object with lots of fields, most of which are set to their default value. There's also a way to pretty-print transactions in a more human-readable, condensed format:
+This command outputs a massive JSON object with many fields, most of which are set to their default value. 
+
+### Transaction output
+
+Instead, it's better to output transactions in a more human-readable, condensed format to print only the fields with actual content:
```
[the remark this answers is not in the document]
barriebyron: there is a lot of legacy "if you do this, it fails" content in the docs... should we remove in this case?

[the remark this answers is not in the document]
barriebyron: I hear ya... and I understand. For context, the other docs content that I considered legacy was the earlier versions of tutorial steps that said "run this command, but it will break" which are now rewritten

docs/zkapps/o1js/interact-with-mina.md
```diff
@@ -81,25 +106,44 @@ Depending on the logic of `myMethod()`, this could print something like the foll
...
 
-As you might have noticed, these account updates weren't created in a very explicit manner. Instead, o1js gives you an imperative API, with "commands" like `state.set()`. Under the hood, these commands create and modify account updates in a transaction, like you saw above. In the end, the entire transaction is sent to the network, as one atomic update. If something fails – for example, one of the account updates has insufficient authorization – the _entire_ transaction is rejected and doesn't get applied. This is in contrast to an EVM contract, where the initial steps of a method call could succeed even if the method fails at a later step.
+In the end, the entire transaction is sent to the network as one atomic update. If something fails – for example, one of the account updates has insufficient authorization – the _entire_ transaction is rejected and doesn't get applied. This is in contrast to an EVM contract, where the initial steps of a method call could succeed even if the method fails at a later step.
 
-## Creating proofs and what they mean
+## Creating proofs
 
-Let's finally see how to create zero-knowledge proofs!
+To create zero-knowledge proofs:
```
barriebyron, line 146:
```suggestion
Finally, here's how to create zero-knowledge proofs!
```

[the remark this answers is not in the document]
barriebyron: We use second person pronoun "you" so the r/w was to avoid using 'we' 
https://github.com/o1-labs/docs2/wiki/Docs-Style-Guide#language-and-grammar 

I've put the fun back in!

docs/zkapps/o1js/interact-with-mina.md
```diff
@@ -112,14 +156,15 @@ const tx = await Mina.transaction(sender, () => {
...
 
-- `MyContract.compile()` creates prover and verification keys from your smart contract.[^1] You need to do this before you can create any proofs!
-- `tx.prove()` goes through your transaction, and creates proofs for all the account updates that came from method calls.
+- `MyContract.compile()` creates prover and verification keys from your smart contract. You must create the keys before you can create any proofs.
 
-[^1]: The name `compile()` is a metaphor for what this function does: creating prover and verifier functions from your code. It doesn't refer to literal "compilation" of JS into a circuit representation. The circuit representation of your code is created by _executing_ it, not by compiling it. Also, the prover function still includes the execution of your JS code as one step.
+  - The name `compile()` is a metaphor for what this function does: creating prover and verifier functions from your code. It doesn't refer to literal "compilation" of JS into a circuit representation. The circuit representation of your code is created by _executing_ it, not by compiling it. Also, the prover function still includes the execution of your JS code as one step.
```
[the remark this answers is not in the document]
barriebyron: Motivation: accessibility, readability. Style guidance https://developers.google.com/style/footnotes 

I agree, let's not bury the super essentials in the ramble. But let's do avoid footnotes. Let me tag you in a bit on an update

(btw, I am really enjoying learning from you in this very thorough doc review, thank you!)

docs/zkapps/o1js/interact-with-mina.md
barriebyron, line 200:
```suggestion
The only valid account updates for a zkApp account are the ones created according to the logic of your smart contract. This core concept is why zkApp smart contracts execute on the client side.
```

docs/zkapps/o1js/interact-with-mina.md
barriebyron, line 236:
```suggestion
MINA amounts, in all o1js APIs and elsewhere in the protocol, are always denominated in nanoMINA = `10^(-9)` MINA, which is why you set `const MINA = 1e9`.
```

docs/zkapps/o1js/interact-with-mina.md
barriebyron, line 240:
```suggestion
The transaction now has three account updates:
```

docs/zkapps/o1js/interact-with-mina.md
```diff
@@ -199,75 +261,59 @@ What's interesting is that the transaction now has 3 account updates:
...
+A key takeaway is: If you want something to become part of your proof, you must put it inside your `@method`.
 
 ## Signing transactions and explicit account updates
 
-Let's recap: We have explained how to write a SmartContract. We've seen how to create a transaction which calls that contract, and how the transaction consists of account updates which were created by o1js under the hood. Now, we'll see an example of creating an account update explicitly. We'll also learn how to use signatures, for authorizing updates to user accounts.
+To recap the workflow covered so far: You write a smart contract, and then create a transaction to call the smart contract. The workflow of this transaction consists of account updates that are created by o1js. 
+
+The next example explicitly creates an account update and shows how to use signatures for authorizing updates to user accounts.
+
+To continue the [Payment to zkApp example](#payment-to-zkapp-example) in the other direction: make a deposit from the user into the zkApp.
```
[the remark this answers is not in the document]
barriebyron: Right, as soon as we focus the content on what the user is doing, we know to speak to them directly "you" (because we are not there with them). It's a thing.

docs/zkapps/o1js/interact-with-mina.md
barriebyron, line 339:
```suggestion
- `AccountUpdate.create()` instantiates this class and attaches the update to the current transaction at the same level where `create` is called. 

    If it is called inside an `@method`, the `AccountUpdate` is created as a child (public input) of the zkApp update.
```

docs/zkapps/o1js/interact-with-mina.md
barriebyron: hmmm... "ripped from content" we don't want that.

I am persuaded

barriebyron, line 421:
```suggestion
```

docs/zkapps/o1js/interact-with-mina.md
```diff
@@ -353,43 +406,53 @@ await tx.prove();
...
+:::
 
-o1js allows you to load and store private and public keys in base58 format. Here's how the sender private key might be created in a script:
+To create the sender private key in a script:
 
 ```ts
 const senderPrivateKey = PrivateKey.fromBase58('EKEQc95...');
-```
 
 In a real server-side deployment, you probably want to load keys from a file or environment variable, instead of hard-coding them in your source code.
```
barriebyron, line 428:
```suggestion
In a real server-side deployment, you probably want to load keys from a file or environment variable, instead of hard-coding them in your source code.
:::
```

docs/zkapps/o1js/interact-with-mina.md
barriebyron: no footnotes, please :)

docs/zkapps/o1js/interact-with-mina.md
```diff
@@ -27,15 +27,36 @@ zkApp programmability is not yet available on the Mina Mainnet. You can get star
...
-You create transactions in o1js by calling `Mina.transaction(...)`, which takes the sender (a public key) and a callback that contains your transaction's logic.
+## Account updates
+
+The fundamental data structure that Mina transactions are built from is called an _account update_. An account update always contains updates to one specific on-chain account. 
+
+For example, if you transfer MINA from one account to another, the balance on two accounts is updated – the sender and the receiver. Therefore, sending MINA requires two account updates. 
+
+Account updates are a flexible and powerful data structure that can express all kinds of updates, events, and preconditions you use to develop smart contracts.
+
+### Transaction JSON object
```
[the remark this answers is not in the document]
barriebyron: good call... isn't it interesting to see how things are consumed by different people!? thanks, coach!

docs/zkapps/o1js/interact-with-mina.md
```diff
@@ -81,25 +106,44 @@ Depending on the logic of `myMethod()`, this could print something like the foll
...
+  - For the `fee`, which you didn't specify, the o1js default value of 0 is used
+  - The `authorization` was filled with a dummy signature.
+    
+In a user-facing zkApp, you typically don't care about setting those values. Instead, you create a transaction like this in the browser and pass it on to the user's wallet. 
+
+The wallet replaces your fee payer account with one that represents the user account, applying the user's settings for the fee. The `authorization` field is filled with a signature created from the user's private key. See [Connecting your zkApp with a user's wallet](/zkapps/how-to-write-a-zkapp-ui#connecting-your-zkapp-with-a-users-wallet).
+
+The second account update has the `'MyContract.myMethod()'` label:
+
+- The update corresponds to the method call. An `@method` call always results in the creation of an account update – an update to the zkApp account itself.
```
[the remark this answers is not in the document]
barriebyron: You are right! https://developers.google.com/style/lists

docs/zkapps/o1js/interact-with-mina.md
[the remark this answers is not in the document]
mitschabaude: I agree that readability would be hurt if the text was littered with footnotes. But I can't see how a single footnote presents a problem in that regard.
This is non-essential info (that's why it's in a footnote), so not being accessible is not really a concern IMO.

docs/zkapps/o1js/interact-with-mina.md
mitschabaude: @barriebyron might this be the reason why it fails to deploy?

docs/zkapps/o1js/interact-with-mina.md
```diff
@@ -199,75 +246,57 @@ What's interesting is that the transaction now has 3 account updates:
...
+In general, there are three kinds of authorizations that an account update can have: a proof, a signature, or none. We'll learn about signatures in the next section. You can find more details in [our docs about permissions](TODO).
 
-<!-- TODO: _ideal_ here would be a little picture of a tree, with a parent and a child node -->
 
-Concretely, in our example, both the zkApp account update and the user account update are public input to the zkApp method call. Intuitively, being public input means that the zkApp can "see" and constrain the update as part of its proof. Here, it means that nobody could change the public key of the receiver, or amount they receive, without making the proof invalid. The update can only contain what the method specified.
+- Proof authorization – used for zkApp accounts when you do an `@method` call. Proofs are verified against the on-chain verification key.
+- Signature authorization – used to update user accounts. Signatures are verified against the account's public key.
+- No authorization – used on updates which don't require authorization. For example, positive balance changes.
```
mitschabaude, line 265: As mentioned [here](https://github.com/o1-labs/docs2/pull/652#discussion_r1361706865) I think we should remove this
```suggestion
```

docs/zkapps/o1js/interact-with-mina.md
barriebyron: you are right! thank you, coach

mitschabaude (approved):
Looks great now! Thanks @barriebyron

PR #669, opened 2023-10-20
Fixed broken link syntax

nicc:
Found a link that wasn't displaying correctly on the actions and reducer page. Fixed it


barriebyron (approved):
thanks for the fix!

PR #675, opened 2023-10-20
Move potential changes to zkProgram into the "experimental" callout

garwalsh:


barriebyron (approved):
thanks for the fix! LGTM

PR #676, opened 2023-10-23
Update GIT hash for the new rampup 6 and remove errors that occurred …

vfrsilva:
…prior to the network started.


barriebyron (approved):
congratulations on genesis!

PR #680, opened 2023-10-24
update tutorial - remove isReady and shutdown

LucidSamuel:
closes https://github.com/o1-labs/docs2/issues/389


barriebyron (approved):
LGTM although I did not test the tutorial with updated code, I trust you, so if you think this is good, then I think it is good :)

PR #681, opened 2023-10-24
closing stale PR

LucidSamuel:
closes https://github.com/o1-labs/docs2/issues/389


barriebyron:
Thanks @LucidSamuel. I thought perhaps it was more involved than just removing the lines. TIL!
Can you also check the code in the other tutorials?

barriebyron (approved):
If you think this is good, then I think this is good.

barriebyron:
@LucidSamuel  can you fix the conflict resolution here?
and verify the code runs?

PR #684, opened 2023-10-26
Fix "zkapps - developers - creating zkapps" page ordering in a sidebar

qwadratic:
image

This PR moves sidebar item "How to write zkApp UI" to the bottom, according to the Creating zkApps articles' logical order (see current ordering on the screenshot above)


barriebyron (approved):
thanks for the update

PR #697, opened 2023-10-27
Make tutorial 8 runnable

qwadratic:
[Tutorial 8](https://docs.minaprotocol.com/zkapps/tutorials/private-inputs-hash-functions) `main.ts` and `mainWhitelist.ts` are not runnable

In `main.ts` there are 2 following errors:
- First one should be fixed by omitting `.init()` transaction as it is being invoked within deploy transaction
- Second one should be fixed by changing send_txn sign() statement from `await send_txn.sign([deployerAccount]).send();` to `await send_txn.sign([deployerAccount, zkAppPrivateKey]).send();` because we want to pay fee from deployer (1st update) and send from zkApp account (2nd update)

Another 3 issues in `mainWhitelist.ts` are following:
- Same issue with explicit `init()` call
- Need to fund fee payer account before minting and sending tokens
- errors related to `requireSignature` usage, which seems to be unneeded as there is no usage of this function in `main.ts` example.

For both contracts, tests have syntax errors in the import statement


Pfed-prog:
omg, thank you so much. This works.

LucidSamuel (approved):
GTG!

PR #700, opened 2023-10-27
fix tutorial 9 typing errors added void type as a second param (where methods don’t return anything, and Field (where a method returns Field)

qwadratic:
When trying to run [Tutorial 9](https://docs.minaprotocol.com/zkapps/tutorials/recursion) sample, the typing error says that `SelfProof` and `Proof` require 2 type arguments (`PublicInput` and `PublicOutput`), so I added `void` type as a second param (where methods don’t return anything, and `Field` (where a method returns `Field`)


iregina:
Amazing! Will test this out today!

examples/zkapps/09-recursion/src/rollup.ts
```diff
@@ -23,14 +13,14 @@ import {
   DeployArgs,
   Proof,
   Permissions,
+  ZkProgram
```
iregina, line 16: Forgot to add this

examples/zkapps/09-recursion/src/rollup.ts
```diff
@@ -1,18 +1,8 @@
...
   Field,
-  Mina,
-  PrivateKey,
-  AccountUpdate,
   SelfProof,
-  Experimental,
   Struct,
-  Bool,
-  Circuit,
-  Poseidon,
```
iregina, line 13: @mitschabaude helped remove all the unused methods etc

iregina (approved):
Looks good! Tested it out and it works. Great job.

PR #703, opened 2023-10-27
Removed line 10 from 'Interact with a smart contract' code snippet

garwalsh:
Line 10 in the first code snippet in the section 'Interact with a smart contract' is blank, so the number 10 is included when you use the copy button


garwalsh:
On further inspection it looks like the blank line 10 is left in there to maintain line numbering with the sample file.

But when you copy and paste from the snippet, you get this:

image

ymekuria (approved):
LGTM!

barriebyron (approved):
another reason we want to omit line numbers, it's always something! thanks for the quick fix

PR #704, opened 2023-10-27
Delete duplicate paragraph

garwalsh:


garwalsh:
@barriebyron it looks like (almost) the same paragraph was repeated twice here.
It looks like I deleted the appropriate paragraph because it has slightly less detail

barriebyron (approved):
good catch!

PR #728, opened 2023-11-14
codebase overview: typo + formatting issue

emberian:
splits the prose back into a paragraph before continuing to the next code block


barriebyron (approved):
approving from syntax point of view, trusting that the codebase technical bits are accurate. Thank you for the PR!

Issue #730, opened 2023-11-16
docs: add FAQs from zkHack


barriebyron:
add https://www.notion.so/o1labs/o1js-ZK-Hack-Office-Hours-Q-A-63c7c09a18044765bd193b4a25f4bc2c?pvs=4

PR #731, opened 2023-11-16
docs: new FAQs from the ZK Hack 

barriebyron:
Closes: #730 

Adds FAQs that came up during zkHack 2023

Question: do we want to include the Mina Protocol FAQs in this doc?

### DIFF
--- a/docs/zkapps/faq.mdx
+++ b/docs/zkapps/faq.mdx
@@ -95,3 +95,66 @@ This example fails for two reasons:
 
 1. `n.toString()` can't be used in circuit code at all. It throws an error during `SmartContract.compile()` because during `compile()`, variables like `n` don't have any JS values attached to them; they represent abstract variables used to build up an abstract arithmetic circuit. So, in general, you can't use any of the methods that read out the JS value of your Field elements: `Field.toString()`, `Field.toBigInt()`, `Bool.toBoolean()` etc.
 2. More subtly, your methods must create the same constraints every time because a proof cannot be verified against a verification key for a differing set of constraints. The code above adds `x.equals(y).assertFalse()` _on condition of_ the value of `n` which leads to constraints varying between executions of the proof.
+
+### Why is the variable currentState set to a value retrieved from the blockchain and then immediately compared to that value?
+
+As represented in the tutorial example code:
+
+```TypeScript
+const currentState = this.num.get();
+
+this.num.assertEquals(currentState);
+```
+
+- The first line of code executes before the proof is generated. 
+- The second line of code creates a precondition that is checked when the proof is sent to the blockchain to be verified. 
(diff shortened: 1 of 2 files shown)


docs/zkapps/faq.mdx
```diff
@@ -95,3 +95,66 @@ This example fails for two reasons:
...
+```TypeScript
+const currentState = this.num.get();
+
+this.num.assertEquals(currentState);
+```
+
+- The first line of code executes before the proof is generated. 
+- The second line of code creates a precondition that is checked when the proof is sent to the blockchain to be verified. 
+
+This ensures that the transaction fails if the value of the field in question has changed.
```
iregina, line 112: Is this another way of saying that that this addresses race conditions vulnerabilities? @jackryanservia @LucidSamuel

docs/zkapps/faq.mdx
```diff
@@ -95,3 +95,66 @@ This example fails for two reasons:
...
+
+### Can I pass hex values into Fields?
+
+Yes, just pass in the appropriate BigInt literal.
+
+`Field(0x1337)`
```
iregina, line 118: Just curious about the context of this question, why wouldn't this be possible in the first place? @jackryanservia @LucidSamuel

iregina:
Left some comments - overall, it looks good just had one question about the Field type questions.

docs/zkapps/faq.mdx
```diff
@@ -95,3 +95,66 @@ This example fails for two reasons:
...
+As represented in the tutorial example code:
+
+```TypeScript
+const currentState = this.num.get();
+
+this.num.assertEquals(currentState);
+```
+
+- The first line of code executes before the proof is generated. 
+- The second line of code creates a precondition that is checked when the proof is sent to the blockchain to be verified.
```
garwalsh, line 110: The second line of code creates a precondition that is checked when the proof is sent in a transaction the blockchain to be verified.

garwalsh:
@barriebyron 
We don't need to draw the distinction between o1js and Mina FAQs here.
It's better framed as zkApp developer questions. They came directly from hackers at ZK Hack

docs/zkapps/faq.mdx
LucidSamuel: It's likely a matter of developer familiarity. When using a new tool, developers often seek clarity, just to be sure, especially in case there are differences in how things work.

docs/zkapps/faq.mdx
barriebyron, line 110:
```suggestion
- The second line of code creates a precondition that is checked when the proof is sent in a transaction to the blockchain to be verified. 
```

garwalsh (approved):

PR #737, opened 2023-11-17
docs: add gadgets to o1js 

barriebyron:
closes https://app.zenhub.com/workspaces/zkapps-product-eng-6130fedb3b0fc600123d8796/issues/gh/o1-labs/docs2/733

### DIFF
--- a/docs/zkapps/o1js/gadgets.mdx
+++ b/docs/zkapps/o1js/gadgets.mdx
@@ -1,0 +1,28 @@
+---
+title: Events
+hide_title: true
+description: How events in zkApps serve as a light-weight way of attaching information about your smart contract execution, use cases of events, and how to declare and emit events.
+keywords:
+  - smart contracts
+  - zkapp
+  - gadgets
+  - bitwise
+  - o1js
+  - blockchain
+---
+
+:::info
+
+Please note that zkApp programmability is not yet available on Mina Mainnet, but zkApps can now be deployed to Berkeley Testnet.
+
+:::
+
+# Gadgets
+
+Gadgets are amalgamations of the lowest level gates in the proof system (addition, multiplication) into more sophisticated boolean operations (XOR, NOT). This simplifies the process of creating new cryptographic primitives. 
+
+In this case, the gadgets form the basis of ECDSA signature verification. These gadgets allow the community to implement other crypto algorithms, such as SHA.
+
+Gadgets.rangeCheck64(): A provable method for efficient 64-bit range checks using lookup tables. (0.14.0, #1181)
+Gadgets.rotate(): A provable method to support bitwise rotation for native field elements. (0.14.0, #1182)
+Gadgets.xor(): A provable method to support bitwise XOR operations for native field elements. (0.14.0, #1177)
--- a/sidebars.js
+++ b/sidebars.js
@@ -60,6 +60,7 @@ module.exports = {
             'zkapps/o1js/on-chain-values',
             'zkapps/o1js/recursion',
             'zkapps/o1js/custom-tokens',
+            'zkapps/o1js/gadgets',
             'zkapps/o1js/merkle-tree',
             'zkapps/o1js/permissions',
             'zkapps/o1js/time-locked-accounts',


docs/zkapps/o1js/gadgets.mdx
```diff
@@ -0,0 +1,100 @@
...
+
+<!--- I'm grabbing context from other sources SWAG, please review carefully--->
+Think of gadgets as atomic logic gates that mask large circuits. 
+
+Gadgets are small, reusable building blocks you can use to create more complex cryptographic constructions. 
+
+Leverage the properties and behavior of these smaller components to build secure and efficient systems.
+
+Questions: 
+- are all gadgets bitwise provers? do they all use ZkProgram? what does that mean? https://github.com/o1-labs/o1js/blob/main/src/examples/zkprogram/gadgets.ts#L18
```
garwalsh, line 32: The example you've linked to is a zkProgram that implements the bitwise gadgets. It shows how a developer would use these gadgets in provable code. So "bitwise prover" is just the name that has been chosen for the proof that is created.

In this example, zkProgram is used, but these gadgets could also be included in a SmartContract

docs/zkapps/o1js/gadgets.mdx
```diff
@@ -0,0 +1,100 @@
+---
+title: Gadgets
```
garwalsh, line 2: @barriebyron 
Check out https://github.com/o1-labs/o1js/blob/main/src/lib/gadgets/gadgets.ts

Lots of copy and examples there to use

docs/zkapps/o1js/gadgets.mdx
```diff
@@ -0,0 +1,100 @@
...
+Gadgets are small, reusable building blocks you can use to create more complex cryptographic constructions. 
+
+Leverage the properties and behavior of these smaller components to build secure and efficient systems.
+
+Questions: 
+- are all gadgets bitwise provers? do they all use ZkProgram? what does that mean? https://github.com/o1-labs/o1js/blob/main/src/examples/zkprogram/gadgets.ts#L18 
+- are all gadgets used/called as a method that moves the result into one of our 8 fields?
+- handle the constant case by using the added functionality in the bindings. For the prover case, how to describe? or go into detail for each gadget?
+- do we want to list gadgets in the gadgets namespace separately from gadgets not in the gadgets namespace? (where are these gadgets?)
+- do we have examples other than https://github.com/o1-labs/o1js/blob/main/src/examples/zkprogram/gadgets.ts?
```
garwalsh, line 36: https://github.com/o1-labs/o1js/blob/main/src/lib/gadgets/gadgets.ts has examples for each gadget

docs/zkapps/o1js/gadgets.mdx
```diff
@@ -0,0 +1,100 @@
...
+Please note that zkApp programmability is not yet available on Mina Mainnet, but zkApps can now be deployed to Berkeley Testnet.
+
+:::
+
+# Gadgets
+
+Gadgets are amalgamations of the lowest level gates in the proof system (addition, multiplication) into more sophisticated boolean operations (XOR, NOT). Gadgets in o1js simplify the process of creating new cryptographic primitives and make it possible to implement other crypto algorithms, such as SHA. 
+
+<!--- I'm grabbing context from other sources SWAG, please review carefully--->
+Think of gadgets as atomic logic gates that mask large circuits.
```
garwalsh, line 25: I think we can remove 25-29

The description above covers it.

docs/zkapps/o1js/gadgets.mdx
```diff
@@ -0,0 +1,100 @@
...
+<!--- I'm grabbing context from other sources SWAG, please review carefully--->
+Think of gadgets as atomic logic gates that mask large circuits. 
+
+Gadgets are small, reusable building blocks you can use to create more complex cryptographic constructions. 
+
+Leverage the properties and behavior of these smaller components to build secure and efficient systems.
+
+Questions: 
+- are all gadgets bitwise provers? do they all use ZkProgram? what does that mean? https://github.com/o1-labs/o1js/blob/main/src/examples/zkprogram/gadgets.ts#L18 
+- are all gadgets used/called as a method that moves the result into one of our 8 fields?
```
garwalsh, line 33: We shouldn't generalise what a gadget returns because by it depends on the gadget and this list will grow

docs/zkapps/o1js/gadgets.mdx
```diff
@@ -0,0 +1,100 @@
...
+
+:::info
+
+Please note that zkApp programmability is not yet available on Mina Mainnet, but zkApps can now be deployed to Berkeley Testnet.
+
+:::
+
+# Gadgets
+
+Gadgets are amalgamations of the lowest level gates in the proof system (addition, multiplication) into more sophisticated boolean operations (XOR, NOT). Gadgets in o1js simplify the process of creating new cryptographic primitives and make it possible to implement other crypto algorithms, such as SHA.
```
garwalsh, line 22: I know I wrote this but I think it needs to be a bit more general.

Gadgets are amalgamations of the lowest level gates in the proof system (addition, multiplication) into more sophisticated operations. Gadgets in o1js simplify the process of creating new cryptographic primitives and make it possible to implement new crypto algorithms, such as SHA.

docs/zkapps/o1js/gadgets.mdx
nicc: Agreed with Gar. I'd also add that fields can contain anything and are not bound to any gadgets. Saving values to them would always be done explicitly

docs/zkapps/o1js/gadgets.mdx
```diff
@@ -0,0 +1,100 @@
...
+
+Leverage the properties and behavior of these smaller components to build secure and efficient systems.
+
+Questions: 
+- are all gadgets bitwise provers? do they all use ZkProgram? what does that mean? https://github.com/o1-labs/o1js/blob/main/src/examples/zkprogram/gadgets.ts#L18 
+- are all gadgets used/called as a method that moves the result into one of our 8 fields?
+- handle the constant case by using the added functionality in the bindings. For the prover case, how to describe? or go into detail for each gadget?
+- do we want to list gadgets in the gadgets namespace separately from gadgets not in the gadgets namespace? (where are these gadgets?)
+- do we have examples other than https://github.com/o1-labs/o1js/blob/main/src/examples/zkprogram/gadgets.ts? 
+- include code snippets?
```
nicc, line 37: I'm happy to sit with you to put together some snippets prior to submitting a PR for review, @barriebyron. Please collate examples from the links Gar supplied as a starting point and let me know when you're ready to review

docs/zkapps/o1js/gadgets.mdx
```diff
@@ -0,0 +1,100 @@
...
+
+Questions: 
+- are all gadgets bitwise provers? do they all use ZkProgram? what does that mean? https://github.com/o1-labs/o1js/blob/main/src/examples/zkprogram/gadgets.ts#L18 
+- are all gadgets used/called as a method that moves the result into one of our 8 fields?
+- handle the constant case by using the added functionality in the bindings. For the prover case, how to describe? or go into detail for each gadget?
+- do we want to list gadgets in the gadgets namespace separately from gadgets not in the gadgets namespace? (where are these gadgets?)
+- do we have examples other than https://github.com/o1-labs/o1js/blob/main/src/examples/zkprogram/gadgets.ts? 
+- include code snippets? 
+- can we need to describe use cases? or just talk about performance gains? or?
+- primitive constraint system, we replaced Poiseidon with gadget? https://github.com/o1-labs/o1js/pull/1177/files#diff-de4ea0a8fb59a20403ab4c9ab9615587c179a8649e1b1b07297f6c829c94aa88L1
```
nicc, line 39: Nope, I wouldn't read into that too much. That's just an import statement. My guess is that Poseidon was imported but never used and that the extraneous import was removed at the same time as the Gadget example was added. git shows one swapped for the other but that does not reflect any deeper code changes

docs/zkapps/o1js/gadgets.mdx
```diff
@@ -0,0 +1,100 @@
...
+- are all gadgets used/called as a method that moves the result into one of our 8 fields?
+- handle the constant case by using the added functionality in the bindings. For the prover case, how to describe? or go into detail for each gadget?
+- do we want to list gadgets in the gadgets namespace separately from gadgets not in the gadgets namespace? (where are these gadgets?)
+- do we have examples other than https://github.com/o1-labs/o1js/blob/main/src/examples/zkprogram/gadgets.ts? 
+- include code snippets? 
+- can we need to describe use cases? or just talk about performance gains? or?
+- primitive constraint system, we replaced Poiseidon with gadget? https://github.com/o1-labs/o1js/pull/1177/files#diff-de4ea0a8fb59a20403ab4c9ab9615587c179a8649e1b1b07297f6c829c94aa88L1 
+- resources, what do we want to provide?
+- why are gadgets listed under [variables](https://docs.minaprotocol.com/zkapps/o1js-reference/modules#variables-1) in the reference docs? is it because a gadget variable is already computed when it is used (so performance is faster?)
```
nicc, line 41: Technically `Gadget` itself is a variable. The reference docs are generated from the code and, since Gadget is defined as an object with function properties, the docs list `Gadget` as a variable. It's just a JavaScript thing. You'll see Poseidon, Permissions and Pickles also listed as variables. This just means that when you refer to those things they are technically defined as variables in the JS runtime. This does not need to be communicated

docs/zkapps/o1js/gadgets.mdx
barriebyron: @garwalsh  can you help me understand this feedback? 
I am not clear what to do with your comment on line 22... 
More general intro about gadgets, you mean? Let's talk so I understand

When we write our code, it turns into gates. Gates are like transistors. Gadgets are built out of gates, but not a program that a user would write.
Gates are configured to perform efficiently.
Gadgets are ... what in o1js?

garwalsh: @barriebyron 
Will use suggestions in future. For this particular piece of feedback I have provided a slightly edited paragraph to replace the one starting at line 22

sidebars.js
```diff
@@ -60,6 +60,17 @@ module.exports = {
             'zkapps/o1js/on-chain-values',
             'zkapps/o1js/recursion',
             'zkapps/o1js/custom-tokens',
+            {
+              type: 'category',
+              label: 'Gadgets',
+              link: {
+                type: 'doc',
+                id: 'zkapps/o1js/gadgets'
```
barriebyron, line 68: @MartinMinkov  can you help me debug this nested category?

docs/zkapps/o1js/bitwise-gadgets.mdx
```diff
@@ -0,0 +1,420 @@
...
+:::info
+
+Please note that zkApp programmability is not yet available on Mina Mainnet, but zkApps can now be deployed to Berkeley Testnet.
+
+:::
+
+# Bitwise Gadgets in o1js
+
+Bitwise operations work on 2-bit patterns of equal lengths by positionally matching their individual bits. 
+
```
barriebyron, line 20: @garwalsh do we want to include the foreign field gadgets? 
(they seem to be different than bitwise)

docs/zkapps/o1js/bitwise-gadgets.mdx
```diff
@@ -0,0 +1,420 @@
...
+- [xor](#xor)
+
+## and
+
+```ts
+and(a: Field, b: Field, length: number) {
+    return and(a, b, length);
+```
+
+The bitwise `and()` gadget on Field elements is equivalent to the [bitwise AND (&)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Bitwise_AND) operator in JavaScript.
```
barriebyron: @garwalsh  I have basic questions:
1. how do we want to "write" the name of the gadget?
For example, I see variations on `and()` or `and` or AND
2. is it a gadget or a gate? we use both terms

docs/zkapps/o1js/bitwise-gadgets.mdx
```diff
@@ -0,0 +1,420 @@
...
+  * `a & b = and`
+
+For details about the implementation, see [AND](https://o1-labs.github.io/proof-systems/specs/kimchi.html?highlight=gates#and) in the Mina book.
+
+The `length` parameter:
+
+- Defines how many bits to compare. 
+- Rounds `length` to the nearest multiple of 16, `paddedLength = ceil(length / 16) * 16`.
+- Constrains both input values to fit into `paddedLength` bits. 
+- Guarantees that the output is has at most `paddedLength` bits.
```
barriebyron: tbh, I couldn't tell if this description applies only to the length parameter or to the and gadget

docs/zkapps/o1js/bitwise-gadgets.mdx
```diff
@@ -0,0 +1,420 @@
...
+- **Note:** Specifying a larger `length` parameter adds additional constraints.
```
barriebyron:
```suggestion
- Specifying a larger `length` parameter adds additional constraints.
```

docs/zkapps/o1js/bitwise-gadgets.mdx
```diff
@@ -0,0 +1,320 @@
...
+
+The bitwise and gadget on Field elements is equivalent to the [bitwise AND (&)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Bitwise_AND) operator in JavaScript. 
+
+- The and gadget works by comparing two bits and returns `1` only if both bits are `1`, otherwise it returns `0`.
+- Constrains both input values to fit into `paddedLength` bits. 
+- Both Field elements must fit into `2^paddedLength - 1` or an error is thrown.
+- Guarantees that the output is has at most `paddedLength` bits.
+- Can be checked by a double generic gate that verifies the following relationship between the values. In the process, it also invokes the [xor()](#xor) gadget that creates additional constraints depending on `length`.
+
+  The generic gate verifies `a + b = sum` and the conjunction equation `2 * and = sum - xor`.
```
barriebyron, line 46: is line 46 here a "double generic gate" follow up? but here we just talk about "generic gate"

docs/zkapps/o1js/bitwise-gadgets.mdx
```diff
@@ -0,0 +1,320 @@
...
+  * `a ^ b = xor`
+  * `a & b = and`
+
+For details about the implementation, see [AND](https://o1-labs.github.io/proof-systems/specs/kimchi.html?highlight=gates#and) in the Mina book.
+
+The `length` parameter:
+
+- Defines how many bits to compare. 
+- Rounds `length` to the nearest multiple of 16, `paddedLength = ceil(length / 16) * 16`.
+- Specifying a larger `length` parameter adds additional constraints.
```
barriebyron, line 59: Two questions here:
1. "adds additional" is redundant. Wdyt about "adds more constraints" or just "adds constraints" (but what are the constraints?)

2. does this consideration on https://github.com/o1-labs/docs2/pull/737/files#diff-c8519360c4a9410ca73564931a0167afc43ba115e125b3719e1c6fa8edc2adc3R305 apply to all gadgets? tradeoffs?

docs/zkapps/o1js/bitwise-gadgets.mdx
```diff
@@ -0,0 +1,320 @@
...
+```ts
+let a = Field(3);    // ... 000011
+let b = Field(5);    // ... 000101
+let c = Gadgets.and(a, b, 2);    // ... 000001
+c.assertEquals(1);
+```
+
+## compactMultiRangeCheck
+
+```ts
```
barriebyron, line 74: we can provide the syntax for each gadget... not sure I have all of them right, but sometimes all a dev needs is the syntax 
(do we want to document the parameters?)

docs/zkapps/o1js/bitwise-gadgets.mdx
```diff
@@ -0,0 +1,320 @@
...
+```
+
+The leftShift gadget is a provable method that supports the bitwise shifting operation that moves the bits of a binary number to the left. Unlike rotation, the bits that fall off at the end are discarded and the vacant positions are filled with zeros.
+
+The leftShift gadget:
+
+- Performs a left shift operation on a Field element.
+- Shifts bits to the left and discards the overflowing bits.
+- Performs these operations with the big-endian 64-bit representation of the number, where the most significant (64th) bit is on the left end and the least significant bit is on the right end.
+- Requires values that are range checked to 64 bits. You can use [rangeCheck64](#rangecheck64).
```
barriebyron, line 111: can use or must use?

docs/zkapps/o1js/bitwise-gadgets.mdx
```diff
@@ -0,0 +1,320 @@
...
+```
+
+The rightShift gadget is a provable method that supports the bitwise shifting operation that moves the bits of a binary number to the right. Unlike [rotate](#rotate), the bits that fall off at the end are discarded and the vacant positions are filled with zeros.
+
+The rightShift gadget:
+
+- Performs a right shift operation on a Field element.
+- Shifts bits to the right and discards the overflowing bits.
+- Performs these operations with the big-endian 64-bit representation of the number, where the most significant (64th) bit is on the left end and the least significant bit is on the right end.
+- Requires values that are range checked to 64 bits. You can use [rangeCheck64](#rangecheck64).
```
barriebyron, line 234: can use or must use?

docs/zkapps/o1js/bitwise-gadgets.mdx
```diff
@@ -0,0 +1,320 @@
...
+
+The rotate gadget is a provable method to support bitwise left or right rotation for native Field elements. 
+
+The rotate gadget:
+
+- Shifts bits to the left or the right. 
+- The rotate left operation (ROL) shifts bits to the left. The bits that fall off the leftmost side wrap around and reappear on the rightmost side.
+- The rotate right operation (ROR) shifts bits to the right. The bits that fall off the rightmost side wrap around and reappear on the leftmost side.
+- Performs these operations with the big-endian 64-bit representation of the number, where the most significant (64th) bit is on the left end and the least significant bit is on the right end.
+- Requires values that are range checked to 64 bits. You can use [rangeCheck64](#rangecheck64).
```
barriebyron, line 265: can use or must use? (asked question already for other gadgets, not sure if the answer if the same for all range checked values)

docs/zkapps/o1js/gadgets.mdx
```diff
@@ -0,0 +1,28 @@
...
+
+:::info
+
+Please note that zkApp programmability is not yet available on Mina Mainnet, but zkApps can now be deployed to Berkeley Testnet.
+
+:::
+
+# Gadgets
+
+Gadgets are small, reusable low-level building blocks that accelerate special computations. Most gadgets build upon custom gates and act as low-level accelerators in the proof system.
```
barriebyron, line 22: I got this description of gadgets from Florian's talk at ZK Hack

docs/zkapps/o1js/bitwise-gadgets.mdx
```diff
@@ -0,0 +1,322 @@
...
+
+:::info
+
+Please note that zkApp programmability is not yet available on Mina Mainnet, but zkApps can now be deployed to Berkeley Testnet.
+
+:::
+
+# Bitwise Gadgets in o1js
+
+Bitwise operations work on 2-bit patterns of equal lengths by positionally matching their individual bits.
```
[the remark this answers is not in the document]
garwalsh: Operations are essentially equivalent to functionality.
e.g. I want to know what x AND y is. To find that out, I need to perform an AND operation.

Gadgets are how o1js facilitates that operation.
So in order to perform an AND operation, I use the AND gadget.

So having thought about this some more, the actual functionality is the primary thing we want to communicate in this page so a more appropriate title for the page is Bitwise Operations.
Then we can also mention that the operations are implemented as Gadgets.

Let's discuss this first so that we are aligned on the messaging.

docs/zkapps/o1js/bitwise-gadgets.mdx
garwalsh: Yes. This should be a separate page "Foreign Field Arithmetic" as it is different functionality

docs/zkapps/o1js/bitwise-gadgets.mdx
garwalsh: "Can use" is fine here and in the other instances.

Using rangecheck before leftshift etc is not a prerequisite.

docs/zkapps/o1js/bitwise-operations.mdx
```diff
@@ -0,0 +1,303 @@
...
+- [rotate()](#rotate)
+- [rangeCheck64()](#rangecheck64)
+- [multiRangeCheck()](#multirangecheck)
+- [compactMultiRangeCheck()](#compactmultirangecheck)
+
+See the type declaration for [Gadgets](/zkapps/o1js-reference/modules#gadgets) in the o1js Reference documentation.
+
+## and()
+
+```ts
```
barriebyron, line 35: I grabbed this "syntax" from http://localhost:3000/zkapps/o1js-reference/modules#gadgets but it's really not a true command syntax, more of an explainer

Did we want to provide literal syntax here? or omit this for each gadget?

docs/zkapps/o1js/bitwise-operations.mdx
```diff
@@ -0,0 +1,303 @@
...
+and(a: Field, b: Field, length: number) => Field
+```
+
+The bitwise `and()` gadget on Field elements is equivalent to the [bitwise AND (&)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Bitwise_AND) operator in JavaScript. 
+
+The `and()` gadget:
```
barriebyron, line 41: @garwalsh  and I talked about how we can "write" gadget names and landed on following how we write about methods, using tick marks and parentheses. Does this Markdown syntax work for the docs?

whatever we do, let's do it consistently throughout

docs/zkapps/o1js/bitwise-operations.mdx
```diff
@@ -0,0 +1,303 @@
...
+Example NOT-ing 4 bits with the checked version using `xor()`:
+
+```ts
+let a = Field(0b0101);
+let b = Gadgets.not(a,4,true);
+
+b.assertEquals(0b1010);
+```
+
+where:
```
barriebyron, line 107: do we need a similar where statement for the examples, or remove this one?

docs/zkapps/o1js/bitwise-operations.mdx
```diff
@@ -0,0 +1,303 @@
...
+
+## rotate()
+
+```ts
+rotate(field: Field, bits: number, direction: 'left' | 'right' = 'left') {
+return rotate(field, bits, direction);
+},
+```
+
+The `rotate()` gadget operates like the [Right shift (>>)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Right_shift) and [Left shift (<<)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Left_shift) operators in JavaScript.
```
barriebyron, line 208: syntax is correct, I think we can ignore the weird highlighting here

docs/zkapps/o1js/gadgets.mdx
```diff
@@ -0,0 +1,24 @@
...
+
+Please note that zkApp programmability is not yet available on Mina Mainnet, but zkApps can now be deployed to Berkeley Testnet.
+
+:::
+
+# Gadgets
+
+Gadgets are small, reusable low-level building blocks that simplify the process of creating new cryptographic primitives. Most gadgets build upon custom gates and act as low-level accelerators in the proof system. 
+
+See [Bitwise Operations](/zkapps/o1js/bitwise-operations).
```
barriebyron, line 24: we'll do the foreign field arthmetic in a new PR

docs/zkapps/o1js/bitwise-operations.mdx
```diff
@@ -0,0 +1,300 @@
...
+
+```ts
+and(a: Field, b: Field, length: number) => Field
+```
+
+The bitwise `and()` gadget on Field elements is equivalent to the [bitwise AND (&)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Bitwise_AND) operator in JavaScript. 
+
+The `and()` gadget:
+
+- Compares two bits and returns 1 only if both bits are 1 and returns 0 if both bits are not 1.
```
nicc, line 43:
```suggestion
- Compares two bits and returns 1 only if both bits are 1. Returns 0 if either bit is not 1.
```

docs/zkapps/o1js/bitwise-operations.mdx
```diff
@@ -0,0 +1,300 @@
+---
+title: Bitwise Gadgets
```
garwalsh, line 2:
```suggestion
title: Bitwise Operations
```

docs/zkapps/o1js/bitwise-operations.mdx
garwalsh: The proposed convention is:
- When referring to the gadgets themselves we use "and()" to reflect how they appear in code
- When referring to a boolean operation we use AND

docs/zkapps/o1js/bitwise-operations.mdx
```diff
@@ -0,0 +1,296 @@
...
+A bitwise NOT operation returns 1 in each bit position if the corresponding bit of the operand is 0 and returns 0 if the corresponding bit of the operand is 1.
+
+The operation:
+
+- Fails if the length or the input value is larger than 254 bits.
+- Operates only over the number of bits specified by the `length` parameter.
+    - The `length` parameter defines how many bits to NOT.
+    - A larger `length` parameter adds more constraints.
+- Is implemented in two different ways:
+    - If the `checked` parameter is set to `false`, it is implemented as a subtraction of the input from the all one bitmask. If the `checked` parameter is not provided, the default implementation uses `false`.
```
nicc, line 84: "from the all one bitmask" seems like an error but I don't know what it should say

docs/zkapps/o1js/bitwise-operations.mdx
```diff
@@ -0,0 +1,296 @@
...
+
+- Takes 4 rows and checks `88*3/4 = 66` bits per row, so is slightly more efficient than [64-bit range checks](#rangecheck64).
+- Supports BigInts up to 264 bits (`3x88-bit`) that supports foreign field multiplication with moduli up to `2^259`.
+- Requires input values less than 88 bits. Throws an error if input values exceed 88 bits.
+
+Example:
+
+```ts
+multiRangeCheck(limbs: Field3) {
+    multiRangeCheck(limbs);
```
nicc, line 270:
```suggestion
  multiRangeCheck(limbs);
```

docs/zkapps/o1js/gadgets.mdx
barriebyron: https://github.com/o1-labs/docs2/pull/774 foreign field PR

docs/zkapps/o1js/bitwise-operations.mdx
```diff
@@ -0,0 +1,456 @@
...
+- [multiRangeCheck()](#multirangecheck)
+- [compactMultiRangeCheck()](#compactmultirangecheck)
+
+## and()
+
+```ts
+and(a: Field, b: Field, length: number) => Field
+```
+
+The bitwise `and()` gadget on Field elements is equivalent to the [bitwise AND (&)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Bitwise_AND) operator in JavaScript.
```
barriebyron, line 43: Is `and()` a provable method? If yes, let's add

docs/zkapps/o1js/bitwise-operations.mdx
```diff
@@ -0,0 +1,456 @@
...
+- Recursively builds a chain of XOR gates. 
+- Verifies that each XOR gate can verify at most 16 bits.
+- Adds another XOR gate to the chain if your input elements exceed 16 bits.
+- Rounds `length` to the nearest multiple of 16, `paddedLength = ceil(length / 16) * 16`.
+- Constrains input values to fit into `paddedLength` bits. 
+- Requires that both Field elements fit into `2^paddedLength`.
+    For example, `length = 2` (`paddedLength = 16`) fails for any input that is larger than `2**16`.
+- Guarantees the output has at most `paddedLength` bits.
+
+The `length` parameter defines how many bits to compare. When choosing the value of the `length` parameter, be sure to carefully weigh the trade-off between an increased number of constraints and security:
```
barriebyron, line 131: we have a variety of ways we talk about the length parameter. Would it make sense to expand the length parameter discussion for the other gadgets to be more like this one?

docs/zkapps/o1js/bitwise-operations.mdx
barriebyron: is and() only for native Field elements? (or maybe all of these bitwise operations are for native field elements?)
I feel like we can clarify

docs/zkapps/o1js/bitwise-operations.mdx
```diff
@@ -0,0 +1,456 @@
...
+
+```ts
+addMod32(a: Field, b: Field) => Field
+```
+
+The `addMod32()` operation adds two Field elements in the range `[0, 2^64]` and returns the result modulo `2^32`. It is a helper method to interact with 3-limb vectors of Fields.
+
+The operation:
+
+- Is addition modulo `2^32`.
```
barriebyron, line 159: Do we need to describe what this is?

docs/zkapps/o1js/gadgets.mdx
```diff
@@ -0,0 +1,28 @@
...
+
+Please note that zkApp programmability is not yet available on Mina Mainnet, but zkApps can now be deployed to Berkeley Testnet.
+
+:::
+
+# Gadgets
+
+Gadgets are small, reusable low-level building blocks that simplify the process of creating new cryptographic primitives. Most gadgets build upon custom gates and act as low-level accelerators in the proof system.
+
+In o1js, you can import these provable and helper methods from the gadget namespace:
```
nicc, line 24:
```suggestion
In o1js, you can import these provable and helper methods from the `Gadgets` namespace:
```

docs/zkapps/o1js/bitwise-operations.mdx
nicc, line 43: Yes I believe it is a provable method, @barriebyron. Here's a suggested edit. Eng. reviewers can confirm but I'm pretty sure.
```suggestion
The bitwise `and()` gadget on Field elements is equivalent to the [bitwise AND (&)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Bitwise_AND) operator in JavaScript. It is a provable method for bitwise and operations.
```

nicc: They are indeed for Field elements. This can be seen in [the function definition](https://github.com/o1-labs/o1js/blob/4efb9e505ba12170cbc10c372815e9532aafc77f/src/lib/gadgets/bitwise.ts#L153). It accepts Field types as inputs. And yes anything provable has to work on Field types so this applies to all of these operations. We only need to reimplement them because they have to be provable. They are well established and already exist in a non-zk context / not on Field types.

docs/zkapps/o1js/bitwise-operations.mdx
nicc: We'd need to confirm whether all length parameters result in additional constraints with larger values. Will defer to eng. review on this one.

docs/zkapps/o1js/bitwise-operations.mdx
mitschabaude: * all of these are provable methods, so maybe good to highlight that more broadly in the introduction, instead of on each method. (re what is it proving - it's proving that you computed the bitwise AND of two numbers)
* all of these operate on native field elements (`Field`). I wouldn't necessarily highlight that on each method because `Field` is kind of the "default type" in provable code

docs/zkapps/o1js/bitwise-operations.mdx
nicc: This is actually (perhaps surprisingly) specific. I suspect that people using this will understand it. afaik it's an efficient way to do addition that is overflow safe. All addition in SHA256 works this way. Will defer to eng. review on whether there's an easy way to explain it here cos that would be nice but I don't think we should go into depth on it here.

docs/zkapps/o1js/bitwise-operations.mdx
```diff
@@ -0,0 +1,456 @@
...
+addMod32(a: Field, b: Field) => Field
+```
+
+The `addMod32()` operation adds two Field elements in the range `[0, 2^64)` and returns the result modulo `2^32`. It is a helper method to interact with 3-limb vectors of Fields.
+
+The operation:
+
+- Is addition modulo `2^32`.
+- Asserts that the result is in the range `[0, 2^32)` using [rangeCheck32()](#rangeCheck32).
+- Uses [divMod32()](#divmod32) internally by adding the two Field elements,  decomposing the result into `remainder` and `quotient`, and returning the `remainder`.
```
mitschabaude, line 161: too much internal detail IMO

mitschabaude (commented):
From reviewing the first half of this, I have a few high-level items of feedback:

* The whole page is awfully close in structure and content to our doccomments. But, the doccomments are supposed to be part of the API reference, so we have them in our docs already. Assuming that we fix the doccomment generation (so that the gadgets commens actually show up in the reference), I'm questioning the value of duplicating this content in such detail.
  * Maybe even the structure of this page should be less like that of an API reference?
* This contains a lot of internal implementation detail. I think that's fine for the doccomments, but mostly overkill here.
* What would be valuable instead is some higher-level "synthesis" of these methods, like a real-world usage example that combines several bitwise gadgets (caveat: that takes time to think up and write of course)

docs/zkapps/o1js/bitwise-operations.mdx
```diff
@@ -0,0 +1,456 @@
...
+- [xor()](#xor)
+- [addMod32()](#addmod32)
+- [divMod32()](#divmod32)
+- [leftShift32()](#leftshift32)
+- [leftShift64()](#leftshift64)
+- [rightShift32()](#rightshift32)
+- [rightShift64()](#rightshift64)
+- [rotate32()](#rotate32)
+- [rotate64()](#rotate64)
+- [rangeCheck32()](#rangecheck32)
```
mitschabaude, line 32: FYI - some of these aren't in the current release as planned, but in this PR: https://github.com/o1-labs/o1js/pull/1259
(but maybe we can merge the PR before release, cc @Trivo25)

docs/zkapps/o1js/bitwise-operations.mdx
```diff
@@ -0,0 +1,456 @@
...
+
+```ts
+xor(a: Field, b: Field, length: number) => Field
+```
+
+The `xor()` gadget is equivalent to the [Bitwise XOR (^)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Bitwise_XOR) operator in JavaScript. It is a provable method to support bitwise XOR operations for native Field elements. 
+
+The XOR operation:
+
+- Compares two bits and returns 1 if two bits differ. Returns 0 if the two bits are equal.
```
nicc, line 122:
```suggestion
- Compares two bits and returns 1 if they differ. Returns 0 if the two bits are equal.
```

docs/zkapps/o1js/bitwise-operations.mdx
```diff
@@ -0,0 +1,456 @@
...
+let a = Field(8n);
+let b = Field(1n << 32n);
+
+Gadgets.addMod32(a, b).assertEquals(Field(8n));
+```
+
+## divMod32()
+
+```ts
+divMod32(a: Field, b: Field) => Field
```
nicc, line 177: As above, could be good to know when this helper is useful.

docs/zkapps/o1js/bitwise-operations.mdx
```diff
@@ -0,0 +1,456 @@
...
+```ts
+const x = Provable.witness(Field, () => Field(0b001100)); // 12 in binary
+const y = Gadgets.leftShift(x, 2); // left shift by 2 bits
+y.assertEquals(0b110000); // 48 in binary
+
+const xLarge = Provable.witness(Field, () => Field(12345678901234567890123456789012345678n));
+leftShift(xLarge, 32); // throws an error since input exceeds 64 bits
+``
+
+## leftShift32()
```
nicc, line 248: This is repeated. Looks like copy-pasta

nicc, line 248:
```suggestion
## rightShift32()
```

Looks like a whole lot of copy-pasta. I'll provide inline suggestions to correct on the assumption that it's the same but right-leaning.

docs/zkapps/o1js/bitwise-operations.mdx
```diff
@@ -0,0 +1,456 @@
...
+
+```ts
+leftShift32(field: Field, bits: number) => Field
```
nicc, line 251:
```suggestion
rightShift32(field: Field, bits: number) => Field
```

docs/zkapps/o1js/bitwise-operations.mdx
```diff
@@ -0,0 +1,456 @@
...
+```
+
+The `leftShift32()` gadget is a provable method that supports the bitwise shifting operation that moves the bits of a binary number to the left. Unlike rotation, the bits that fall off at the end are discarded and the vacant positions are filled with zeros.
+
+The operation:
+
+- Performs a left shift operation on a Field element.
```
nicc, line 258:
```suggestion
- Performs a right shift operation on a Field element.
```

docs/zkapps/o1js/bitwise-operations.mdx
Trivo25: The PR is ready from my side, except two or three tiny follow ups - I am just waiting on the crypto stamp, unless you think thats not needed

docs/zkapps/o1js/bitwise-operations.mdx
```diff
@@ -0,0 +1,345 @@
...
+c.assertEquals(1);
+```
+
+## not()
+
+```ts
+not(a: Field, length: number, checked: boolean) => Field
+```
+
+The bitwise `not()` gadget is a provable equivalent to the [Bitwise NOT (~)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Bitwise_NOT) operator in JavaScript. It receives a `Field` element and negates each bit of its binary representation, turning all the `1`s into `0`s and all the `0`s into `1`s. It essentially flips all the bits in a `Field` element. This results in a new binary number which is returned as a `Field` element.
```
barriebyron, line 71: so we want to use the tick marks for 1s and 0s here? Seems like we can use plain text 

(love the r/w @nicc !)

docs/zkapps/o1js/bitwise-operations.mdx
```diff
@@ -0,0 +1,345 @@
...
+
+## leftShift32()
+
+```ts
+leftShift32(field: Field, bits: number) => Field
+```
+
+The `leftShift32()` gadget is a provable equivalent to the [Left shift (<<)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Left_shift) operator in JavaScript. It moves the bits of a binary number to the left by the specified number of `bits`. Any bits that fall off the left side are discarded. `0`s are padded in from the right. It returns a new `Field` element that is range checked to 32 bits.
+
+The input `Field` should not 32 bits in size. You can use [rangeCheck32](#rangecheck32) to ensure this.
```
barriebyron, line 131:
```suggestion
The input `Field` must not exceed 32 bits in size. You can use [rangeCheck32](#rangecheck32) to ensure this.
```

docs/zkapps/o1js/bitwise-operations.mdx
```diff
@@ -0,0 +1,345 @@
...
+
+## leftShift64()
+
+```ts
+leftShift64(field: Field, bits: number) => Field
+```
+
+The `leftShift64()` gadget is a provable equivalent to the [Left shift (<<)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Left_shift) operator in JavaScript. It moves the bits of a binary number to the left by the specified number of `bits`. Any bits that fall off the left side are discarded. `0`s are padded in from the right. It returns a new `Field` element that is range checked to 64 bits.
+
+The input `Field` should not 64 bits in size. You can use [rangeCheck64](#rangecheck64) to ensure this.
```
barriebyron, line 149:
```suggestion
The input `Field` must not exceed 64 bits in size. You can use [rangeCheck64](#rangecheck64) to ensure this.
```

docs/zkapps/o1js/bitwise-operations.mdx
```diff
@@ -0,0 +1,345 @@
...
+
+## rightShift64()
+
+```ts
+rightShift64(field: Field, bits: number) => Field
+```
+
+The `rightShift64()` gadget is a provable equivalent to the [Right shift (>>)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Right_shift) operator in JavaScript. It moves the bits of a binary number to the right by the specified number of `bits`. Any bits that fall off the right side are discarded. `0`s are padded in from the left. It returns a new `Field` element.
+
+The input `Field` should not 64 bits in size. You can use [rangeCheck64](#rangecheck64) to ensure this.
```
barriebyron, line 170:
```suggestion
The input `Field` must not exceed 64 bits in size. You can use [rangeCheck64](#rangecheck64) to ensure this.
```

same comment as above
"must not" or "cannot" (but not should)

let's add "exceeds"

docs/zkapps/o1js/bitwise-operations.mdx
```diff
@@ -0,0 +1,345 @@
...
+
+## rotate64()
+
+```ts
+rotate64(field: Field, bits: number, direction: 'left' | 'right' = 'left') {
+  return rotate64(field, bits, direction);
+},
+```
+
+The `rotate64()` gadget performs provable bit rotation on 32 bit numbers. It is similar to left- and right shift, except the bits that fall off the end will wrap around to reappear on the opposite side instead of being discarded. It accepts a `Field` element, the number of `bits` to rotate, and a `direction` of 'left' or 'right'. The default direction is 'left'.
```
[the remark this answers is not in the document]
barriebyron: to improve readability, I think plain text works for the description. If we want to use code literals, then let's use tick marks

docs/zkapps/o1js/bitwise-operations.mdx
```diff
@@ -0,0 +1,344 @@
...
+
+:::info
+
+Please note that zkApp programmability is not yet available on Mina Mainnet, but zkApps can now be deployed to Berkeley Testnet.
+
+:::
+
+# Bitwise Operations
+
+Bitwise operations manipulate individual bits within a binary represenbtation of a number. They can at times resemble boolean operations but apply to a sequence of bits instead of booleans. Bitwise operations are generally available in most programming languages, including TypeScript. o1js provides versions of them that operate on `Field` elements and result in the necessary circuit constraints to generate a zero knowledge proof of the computation. This is especially useful when implementing hashing algorithms such as SHA256.
```
garwalsh, line 19:
```suggestion
Bitwise operations manipulate individual bits within a binary representation of a number. They can at times resemble boolean operations but apply to a sequence of bits instead of booleans. Bitwise operations are generally available in most programming languages, including TypeScript. o1js provides versions of them that operate on `Field` elements and result in the necessary circuit constraints to generate a zero knowledge proof of the computation. This is especially useful when implementing hashing algorithms such as SHA256.
```

docs/zkapps/o1js/bitwise-operations.mdx
```diff
@@ -0,0 +1,345 @@
...
+
+## addMod32()
+
+```ts
+addMod32(a: Field, b: Field) => Field
+```
+
+The `addMod32()` helper performs addition that will overflow on 32 bit numbers, much like the `int32` type. It returns the result of addition modulo `2^32` in a new `Field` element.
+
+The input `Field`s should not 64 bits in size. You can use [rangeCheck64](#rangecheck64) to ensure this.
```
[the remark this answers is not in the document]
mitschabaude: changing to rangeCheck32 which is correct
```suggestion
The input `Field`s must not exceed 32 bits in size. You can use [rangeCheck32](#rangecheck32) to ensure this.
```

docs/zkapps/o1js/bitwise-operations.mdx
```diff
@@ -0,0 +1,344 @@
...
+let a = Field(8n);
+let b = Field(1n << 32n);
+
+Gadgets.addMod32(a, b).assertEquals(Field(8n));
+```
+
+## divMod32()
+
+```ts
+divMod32(field: Field) => [Field, Field]
```
mitschabaude, line 259:
```suggestion
divMod32(field: Field) => { remainder: Field, quotient: Field }
```

mitschabaude (approved):
I applied @barriebyron's suggestions and one more. Good to go from my side!

Issue #739, opened 2023-11-21
Doc Update: Tutorial 09 - Recursion gives error because of publicInputType. 


LuffySama-Dev:
![recurrsionTypeIssue01](https://github.com/o1-labs/docs2/assets/55076843/9131a486-6812-4b26-99d7-2e43f0475305)
![recurrsionTypeIssue02](https://github.com/o1-labs/docs2/assets/55076843/5df384ef-6a5b-4fb5-8892-108db95d7ce2)

![recurrsionTypeIssue03](https://github.com/o1-labs/docs2/assets/55076843/7eb7392f-4899-4336-b395-2e5f736af79b)
![recurrsionTypeIssue04](https://github.com/o1-labs/docs2/assets/55076843/bbe5ec48-0926-4657-904b-acba49b33b8c)

In above four images, if you will see we are getting the type error in method as the type of publicInput is not matching to the state.

I am creating a pull request for this issue.

Saurabhpatil-dev:
Hi @barriebyron ,

I have created a Pull request for this : #740

Saurabhpatil-dev:
Also, I went through the main.ts file mentioned in the documentation and seems we are using old version of o1js 0.12.* .

barriebyron:
@Saurabhpatil-dev  yes, please get the latest version and do your awesome updates, thank you!

Issue #735, opened 2023-11-16
main.ts in Tutorial 1 fails CI, tutorial runner 


barriebyron:
https://github.com/o1-labs/docs2/actions/runs/6895347054/job/18759106952?pr=731

The tutorial runner for main.ts in tutorial 1 is broken

Martin suggests: maybe the tutorial runner is placing code blocks in a wrong way and causing this error

Docs infra changes will benefit from including Martin or Florian on PR reviews so that we don't break anything

Saurabhpatil-dev:
Hi @barriebyron ,

I think I know why the build is failing. 
I am adding the screen shots and will explain them. 
Hope you find this helpful. 
I will add a PR for this 😃.

> Okay, in below image if you will see **we have already declared the publickey and private key for Deployer and Sender.**
> 
> **Line number 13-14:**

![Line 13 -14](https://github.com/o1-labs/docs2/assets/55076843/aa4e1757-d89f-4972-a1ad-742aa10f09a1)



> **Now, see here below. We are redeclaring it which is not required. Hence giving us error.**
> 
> **Line number 19-20:**

![line 19-20](https://github.com/o1-labs/docs2/assets/55076843/96c8e0b7-d765-4d59-adec-91cc8d39fed1)
 


Solution will be to remove the code on the line number 19-20 and see if there are any other similar issue further.

dkijania:
Closing as obsolete: the tutorial runner this issue targets no longer exists.

`scripts/tutorial-runner.ts` was replaced by the examples matrix in `70915b36` ("ci: add tutorial tests", Oct 2024). `test-tutorials.yml` now builds and executes the checked-in projects under `examples/zkapps/` directly, so nothing invokes the runner — `grep -rn tutorial-runner` across `package.json`, `.github/`, `docs/` and `scripts/` returns only the orphaned script itself. Several issues in this cluster also link the archived `o1-labs/docs2` wiki.

What this cluster asked for — automated proof that each tutorial's code actually works — is being carried forward in **#1249**: an example project for every tutorial, every project built *and run* rather than build-only, real tests in the nine packages that ship an empty jest harness, CI on pull requests instead of only after merge, and a doc↔code drift check that replaces the runner's line-number coupling.

If something specific here is not covered by #1249, please say so on that issue and it will be added to a phase.

PR #740, opened 2023-11-21
:memo: Updating Tutorial-09, Tutorial-01 and Recursion page to resolve the issue #739 , #735 , #736 . Thanks.

LuffySama-Dev:
Solution for issue #739 

recurrsionTypeIssue01
recurrsionTypeIssue02

In above two images, if you will see we are getting the type error in method as the type of publicInput is not matching to the state.

So, I am just modifying the publicInputType to publicInput similar to the previous versions. You can check the similar fix in previous commit here:  #738


Saurabhpatil-dev:
Hi @barriebyron ,

Sorry for adding multiple issue in one pull request. 
But these are all working. 
Let me know how I can help to keep them clean next time ?

Issue which are also fixed:
Tutorial runner work locally #736
main.ts in Tutorial 1 fails CI, tutorial runner #735

docs/zkapps/tutorials/01-hello-world.mdx
```diff
@@ -379,8 +379,8 @@ All smart contracts that you create with the zkApp CLI use similar code:
 Comments break down each stage:
 
 ```ts src/main.ts
-19 const { privateKey: deployerKey, publicKey: deployerAccount } = Local.testAccounts[0];
-20 const { privateKey: senderKey, publicKey: senderAccount } = Local.testAccounts[1];
+19
```
barriebyron, line 382: @Saurabhpatil-dev  
To verify this code update, please make the corresponding change in the example code file https://github.com/o1-labs/docs2/blob/main/examples/zkapps/01-hello-world/src/main.ts

Saurabhpatil-dev: Hi @barriebyron ,

I compared both of the files and it seems the one which are asking me to update is already correct and matching to the changes I committed.

I am attaching the screenshot for your reference:

The file on **_left side is the one I have committed_** and the **_another one which you are asking me to update is on right side_**.  
The duplicated code was removed by me from the left file on line number 19-20. 
But when checked on the file in right side the duplicate code is already removed. So, I don't think so we need to update it. 
Please let me know if you would like me to update the file to match the line numbers in both files or make any other changes.

Screenshot 2023-11-22 at 10 32 27 PM

Thank you 😃

barriebyron (approved):
thanks @Saurabhpatil-dev for these fixes!

PR #775, opened 2023-12-06
fix: mistaken parameter & event names

BerzanXYZ:
Parameter name `id` is used as `userId`, and event name `verified` is used as `id`.

This change fixes it.

https://docs.minaprotocol.com/zkapps/tutorials/oracle#define-the-verify-method


barriebyron (approved):
thanks for the fix contribution

PR #774, opened 2023-12-05
docs: add foreign field arithmetic to o1js

barriebyron:
This PR closes https://app.zenhub.com/workspaces/sdk-6130fedb3b0fc600123d8796/issues/gh/o1-labs/docs2/766 

Foreign field arithmetic preview https://docs2-git-ff-arithmetic-minadocs.vercel.app/zkapps/o1js/foreign-fields

This PR adds the required foreign field classes to the API reference  docs/zkapps/o1js-reference/modules.md

PR #777, opened 2023-12-07
Foreign fields pt 2

mitschabaude:
Additional content about foreign fields on top of #774

TODO:
* [x] build API reference for `ForeignField` and update links that should point to it

### DIFF
--- a/docs/zkapps/o1js-reference/classes/Unconstrained.md
+++ b/docs/zkapps/o1js-reference/classes/Unconstrained.md
@@ -6,8 +6,8 @@ Container which holds an unconstrained value. This can be used to pass values
 between the out-of-circuit blocks in provable code.
 
 Invariants:
-- An `Unconstrained`'s value can only be accessed in auxiliary contexts.
-- An `Unconstrained` can be empty when compiling, but never empty when running as the prover.
+- An `Unconstrained`'s value can be accessed only in auxiliary contexts.
+- An `Unconstrained` container can be empty when compiling, but never empty when running as the prover.
   (there is no way to create an empty `Unconstrained` in the prover)
 
 **`Example`**
@@ -76,7 +76,7 @@ class MyContract extends SmartContract {
 
 #### Defined in
 
-[lib/circuit_value.ts:514](https://github.com/o1-labs/snarkyjs/blob/9d34c7fcf/src/lib/circuit_value.ts#L514)
+[lib/circuit_value.ts:514](https://github.com/o1-labs/o1js/blob/9d34c7fcf/src/lib/circuit_value.ts#L514)
 
 ## Properties
 
--- a/docs/zkapps/o1js-reference/classes/Unconstrained.md
+++ b/docs/zkapps/o1js-reference/classes/Unconstrained.md
@@ -1,0 +1,195 @@
+[o1js](../README.md) / [Exports](../modules.md) / Unconstrained
+
+# Class: Unconstrained<T\>
+
+Container which holds an unconstrained value. This can be used to pass values
+between the out-of-circuit blocks in provable code.
+
+Invariants:
+- An `Unconstrained`'s value can only be accessed in auxiliary contexts.
+- An `Unconstrained` can be empty when compiling, but never empty when running as the prover.
+  (there is no way to create an empty `Unconstrained` in the prover)
+
+**`Example`**
+
+```ts
+let x = Unconstrained.from(0n);
+
+class MyContract extends SmartContract {
+  `@method` myMethod(x: Unconstrained<bigint>) {
+
+    Provable.witness(Field, () => {
+      // we can access and modify `x` here
+      let newValue = x.get() + otherField.toBigInt();
+      x.set(newValue);
+
+      // ...
+    });
+
+    // throws an error!
+    x.get();
+  }
+```
+
+## Type parameters
+
+| Name |
+| :------ |
+| `T` |
+
+## Table of contents
+
+### Constructors
+
+- [constructor](Unconstrained.md#constructor)
+
+### Properties
+
+- [option](Unconstrained.md#option)
+- [provable](Unconstrained.md#provable)
+
+### Methods
+
+- [get](Unconstrained.md#get)
+- [set](Unconstrained.md#set)
+- [from](Unconstrained.md#from)
+- [witness](Unconstrained.md#witness)
+
+## Constructors
+
+### constructor
+
+• `Private` **new Unconstrained**<`T`\>(`isSome`, `value?`)
+
+#### Type parameters
+
+| Name |
+| :------ |
+| `T` |
+
+#### Parameters
+
+| Name | Type |
+| :------ | :------ |
+| `isSome` | `boolean` |
+| `value?` | `T` |
+
+#### Defined in
+
+[lib/circuit_value.ts:514](https://github.com/o1-labs/snarkyjs/blob/9d34c7fcf/src/lib/circuit_value.ts#L514)
+
+## Properties
+
+### option
+
+• `Private` **option**: { `isSome`: ``true`` ; `value`: `T`  } \| { `isSome`: ``false`` ; `value`: `undefined`  }
+
+#### Defined in
+
+[lib/circuit_value.ts:510](https://github.com/o1-labs/snarkyjs/blob/9d34c7fcf/src/lib/circuit_value.ts#L510)
+
+___
+
+### provable
+
+▪ `Static` **provable**: [`Provable`](../modules.md#provable-1)<[`Unconstrained`](Unconstrained.md)<`any`\>\>
+
+#### Defined in
+
+[lib/circuit_value.ts:558](https://github.com/o1-labs/snarkyjs/blob/9d34c7fcf/src/lib/circuit_value.ts#L558)
+
+## Methods
+
+### get
+
+▸ **get**(): `T`
+
+Read an unconstrained value.
+
+Note: Can only be called outside provable code.
+
+#### Returns
+
+`T`
+
+#### Defined in
+
+[lib/circuit_value.ts:523](https://github.com/o1-labs/snarkyjs/blob/9d34c7fcf/src/lib/circuit_value.ts#L523)
+
+___
+
+### set
+
+▸ **set**(`value`): `void`
+
+Modify the unconstrained value.
+
+#### Parameters
+
+| Name | Type |
+| :------ | :------ |
+| `value` | `T` |
+
+#### Returns
+
+`void`
+
+#### Defined in
+
+[lib/circuit_value.ts:537](https://github.com/o1-labs/snarkyjs/blob/9d34c7fcf/src/lib/circuit_value.ts#L537)
+
+___
+
+### from
+
+▸ `Static` **from**<`T`\>(`value`): [`Unconstrained`](Unconstrained.md)<`T`\>
+
+Create an `Unconstrained` with the given `value`.
+
+#### Type parameters
+
+| Name |
+| :------ |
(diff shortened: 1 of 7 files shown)


nicc (approved):

docs/zkapps/o1js-reference/classes/Unconstrained.md
```diff
@@ -0,0 +1,195 @@
+[o1js](../README.md) / [Exports](../modules.md) / Unconstrained
+
+# Class: Unconstrained<T\>
+
+Container which holds an unconstrained value. This can be used to pass values
+between the out-of-circuit blocks in provable code.
+
+Invariants:
+- An `Unconstrained`'s value can only be accessed in auxiliary contexts.
```
barriebyron, line 9:
```suggestion
- An `Unconstrained`'s value can be accessed only in auxiliary contexts.
```

docs/zkapps/o1js-reference/classes/Unconstrained.md
```diff
@@ -0,0 +1,195 @@
+[o1js](../README.md) / [Exports](../modules.md) / Unconstrained
+
+# Class: Unconstrained<T\>
+
+Container which holds an unconstrained value. This can be used to pass values
```
barriebyron, line 5:
```suggestion
Container which holds an unconstrained value. You can use this to pass values
```

barriebyron: I don't know if we want to make the API docs actionable, but if we did, this would be a valid suggestion

docs/zkapps/o1js/foreign-fields.mdx
```diff
@@ -108,6 +106,214 @@ let y = SmallField.from(5n); // convert from bigint or number
 y.toBigInt() === 5n; // convert to bigint
 ```
 
-<!-- TODO replace with link to /ForeignField when it exists -->
+As usual, you can find more information about each method in the [API reference](../o1js-reference/classes/ForeignField).
+
+## Three kinds of foreign field
```
barriebyron, line 111:
```suggestion
## Three kinds of foreign fields
```

docs/zkapps/o1js/foreign-fields.mdx
```diff
@@ -108,6 +106,214 @@ let y = SmallField.from(5n); // convert from bigint or number
...
+
+If the basic usage examples look straightforward, here is where it gets a bit complicated.
+
+For each `ForeignField` class created with `createForeignField()`, there are actually three different variants. We call them _unreduced_, _almost reduced_ and _canonical_.
```
barriebyron, line 115:
```suggestion
For each `ForeignField` class created with `createForeignField()`, there are actually three different variants: _unreduced_, _almost reduced_ and _canonical_.
```

docs/zkapps/o1js/foreign-fields.mdx
```diff
@@ -108,6 +106,214 @@ let y = SmallField.from(5n); // convert from bigint or number
...
+Unreduced field elements just have the `ForeignField` type. For the other two variants, there are narrower base types that are common to each variant:
+
+```ts
+import { AlmostReducedField, CanonicalField } from 'o1js';
+
+y satisfies AlmostReducedField;
+z satisfies CanonicalField;
+```
+
+In the following, we explain when to use the different variants, and how to convert between them. You don't need to learn all of it, though: The type system will guide you to use the right variant in each situation.
```
barriebyron, line 134:
```suggestion
In the following section, you learn when to use the different variants and how to convert between them. You don't need to learn all of it, though: The type system guides you to use the right variant in each situation.
```

docs/zkapps/o1js/foreign-fields.mdx
```diff
@@ -108,6 +106,214 @@ let y = SmallField.from(5n); // convert from bigint or number
...
+```
+
+In short, **unreduced** means that a value can be larger than the modulus.
+
+For example, if `x` has the value 16, it is valid for `x.add(x)` to contain the value 32. The addition is correct modulo 17, but doesn't guarantee a result smaller than 17.
+
+:::note
+
+Unreduced doesn't usually mean that the underlying witness is larger than the modulus. It just means that we have not _proved_ it to be smaller.
+A malicious prover _could_ make it larger, by slightly modifying their local version of o1js and creating a proof with that.
```
barriebyron, line 154:
```suggestion
Unreduced doesn't usually mean that the underlying witness is larger than the modulus. It just means that it is not _proved_ to be smaller.
A malicious prover _could_ make it larger by slightly modifying their local version of o1js and creating a proof with that version.
```

docs/zkapps/o1js/foreign-fields.mdx
```diff
@@ -108,6 +106,214 @@ let y = SmallField.from(5n); // convert from bigint or number
...
+let zz = zAlmost.mul(zAlmost); // zAlmost.mul() is defined
+
+// but .mul() returns an unreduced field again:
+assert(zz instanceof SmallField.Unreduced);
+
+// zAlmost.inv() is defined, and returns an almost reduced field:
+assert(zAlmost.inv() instanceof SmallField.AlmostReduced);
+```
+
+It can be convenient to require almost reduced fields as inputs to your smart contract. To do so, create a class which can also serve as a type, and use its `.provable` property when passing to the state decorator:
```
barriebyron, line 189:
```suggestion
It can be convenient to require almost reduced fields as inputs to your smart contract. To do that, create a class that can also serve as a type and use its `.provable` property when passing to the state decorator:
```

docs/zkapps/o1js/foreign-fields.mdx
```diff
@@ -108,6 +106,214 @@ let y = SmallField.from(5n); // convert from bigint or number
...
+    this.x.set(x.assertAlmostReduced());
+  }
+}
+```
+
+#### What does almost reduced mean?
+
+The definition of almost reduced is somewhat technical. The main motivation is to guarantee that the way we prove modular multiplication is sound. That is definitely true for field elements `< 2^259`. (Recall that we require the modulus to be `< 2^259`.)
+
+However, we actually prove a stronger condition, which lets us save a few constraints in some places:
```
barriebyron, line 208:
```suggestion
The definition of almost reduced is somewhat technical. The main motivation is to guarantee that the way you prove modular multiplication is sound. That is definitely true for field elements `< 2^259`. (Recall that the modulus is required to be `< 2^259`.)

However, we actually prove a stronger condition, which lets us save a few constraints in some places:
```

docs/zkapps/o1js/foreign-fields.mdx
```diff
@@ -108,6 +106,214 @@ let y = SmallField.from(5n); // convert from bigint or number
...
+
+:::note
+
+Example: Assume `x` is a `UInt256` holding the value `2^130`. After computing `z = x.mul(x)`, it is valid for `z` to be `2^260`.
+
+However, by calling `z.assertAlmostReduced()`, we prove that `z` is smaller than `2^259` and safe to use in another multiplication. According to our stronger definition, we even have `z < 2^256`.
+
+:::
+
+Why are we exposing `AlmostReducedField` as a separate type, and don't _always_ prove conditions necessary for multiplication? Because that would take up additional constraints!
```
barriebyron, line 220:
```suggestion
However, by calling `z.assertAlmostReduced()`, you prove that `z` is smaller than `2^259` and is safe to use in another multiplication. According to this stronger definition, you even have `z < 2^256`.

:::

Why is `AlmostReducedField` exposed as a separate type, and why is _always_ prove conditions required for multiplication? Because that would take up additional constraints!
```

docs/zkapps/o1js/foreign-fields.mdx
```diff
@@ -108,6 +106,214 @@ let y = SmallField.from(5n); // convert from bigint or number
...
+
+`ForeignField` is built to allow you to use the minimum amount of constraints, in a way that is safely guided by the type system. See the section below on [minimizing constraints](#minimizing-constraints) for more details.
```
barriebyron, line 222:
```suggestion
`ForeignField` is built to allow you to use the minimum amount of constraints in a way that is safely guided by the type system. See [minimizing constraints](#minimizing-constraints) for more details.
```

docs/zkapps/o1js/foreign-fields.mdx
```diff
@@ -108,6 +106,214 @@ let y = SmallField.from(5n); // convert from bigint or number
...
+
+An operation that is only possible on canonical fields is the boolean equality check:
+
+```ts
+let xCanonical = x.assertCanonical();
+let yCanonical = y.assertCanonical();
+let isEqual = xCanonical.equals(yCanonical);
+```
+
+We require inputs to be canonical for this because we check for strict equality, not equality modulo the field size. Note that being strictly unequal does not imply being unequal as field elements, so `equals()` on non-canonical fields would be error-prone.
```
barriebyron, line 268:
```suggestion
Inputs must be canonical for boolean equality check operations because the operation checks for strict equality, not equality modulo the field size. Note that being strictly unequal does not imply being unequal as field elements, so `equals()` on non-canonical fields would be error-prone.
```

docs/zkapps/o1js/foreign-fields.mdx
```diff
@@ -108,6 +106,214 @@ let y = SmallField.from(5n); // convert from bigint or number
...
+
+`assertAlmostReduced()` takes any number of inputs, but is by far the most efficient with multiples of 3. For example:
+
+- 1 input takes 4.5 constraints
+- 2 inputs take 5 constraints
+- 3 inputs take 5.5 constraints
+
+#### `sum()`
+
+Another opportunity to save constraints is when doing many additions or subtractions in a row. Instead of doing something like `x.add(y).sub(z)`, use `ForeignField.sum()`:
```
barriebyron, line 297:
```suggestion
Another opportunity to save constraints is when many additions or subtractions are performed in a row. Instead of doing something like `x.add(y).sub(z)`, use `ForeignField.sum()`:
```

barriebyron (approved):
@mitschabaude  see the minor update suggestions.
Our docs style is to use the second person pronoun "you" and avoid using third person "we" 
https://developers.google.com/style/person

docs/zkapps/o1js/foreign-fields.mdx
mitschabaude:
```suggestion
In the following section, you learn when to use the different variants, and how to convert between them. You don't need to remember all of it, though: The type system guides you to use the right variant in each situation.
```

docs/zkapps/o1js/foreign-fields.mdx
mitschabaude:
```suggestion
Inputs must be canonical for  `equals()` because the operation checks for strict equality, not equality modulo the field size. Note that being strictly unequal does not imply being unequal as field elements, so `equals()` on non-canonical fields would be error-prone.
```
Trying to be a bit less verbose

docs/zkapps/o1js/foreign-fields.mdx
mitschabaude: This is a tough one for me, because "we" is so standard in the mathematical arguments I'm used to writing

```suggestion
However, by calling `z.assertAlmostReduced()`, you prove that `z` is smaller than `2^259` and safe to use in another multiplication. According to the stronger definition, you even have `z < 2^256`.

:::

Why is `AlmostReducedField` exposed as a separate type, instead of _always_ proving conditions necessary for multiplication? Because that would take up additional constraints!
```

mitschabaude: "necessary condition" is standard math language so I'm going to leave that

PR #785, opened 2023-12-15
ECDSA usage overview

Trivo25:
https://docs2-git-feature-ecdsa-overview-minadocs.vercel.app/zkapps/o1js/ecdsa Adds an overview of how to use ECDSA.

There will be a separate in-depth tutorial that demonstrates how to use ECDSA and Keccak to verify Ethereum data.

### DIFF
--- a/docs/zkapps/o1js/ecdsa.mdx
+++ b/docs/zkapps/o1js/ecdsa.mdx
@@ -22,31 +22,33 @@ Please note that zkApp programmability is not yet available on Mina Mainnet, but
 # ECDSA
 
 ECDSA, or Elliptic Curve Digital Signature Algorithm, is a cryptographic algorithm used to sign and verify messages. It is used in many blockchains, including Ethereum, to sign transactions.
-ECDSA works with different elliptic curves but most commonly with the secp256k1 curve. This curve is also used in Bitcoin and Ethereum.
+ECDSA works with different elliptic curves but most commonly with the [secp256k1](/glossary#secp256k1) curve. This curve is also used in Bitcoin and Ethereum.
 
-# Why ECDSA?
+## Why ECDSA?
 
-In order to interact with other blockchains or verify data from the outside world, o1js needs to be able to verify signatures.
-ECDSA is a widely used algorithm that is supported by many libraries and tools. It is also used in Ethereum, so it is important to support it in o1js as well. For example, Ethereum transactions are signed using ECDSA over the secp256k1 curve.
-If a zkApp developer wanted to verify an Ethereum transaction and make a statement about it, they would need to be able to verify the signature of the transaction. This is why ECDSA is important for zkApps.
+To interact with other blockchains and verify data from the outside world, o1js needs to be able to verify signatures.
+ECDSA is a widely used algorithm that is supported by many libraries and tools. For example, Ethereum transactions are signed using ECDSA over the secp256k1 curve.
+As a zkApp developer, when you want to verify an Ethereum transaction and make a statement about it, you must be able to verify the signature of the transaction which is why ECDSA is important for zkApps.
 
-# Basic usage
+## Basic usage
 
-The ECDSA gadget is used to verify ECDSA signatures. It takes as input the message, the signature, and the public key of the signer. It outputs a `Bool` indicating whether the signature is valid or not.
+The ECDSA gadget is used to verify ECDSA signatures. The gadget takes as input the message, the signature, and the public key of the signer. It outputs a `Bool` indicating whether the signature is valid.
 
-Before we can verify a signature, we must initiate the gadget with a curve configuration. The most commonly used curve is secp256k1, which is also used in Ethereum. We can initiate the curve as follows:
+Before you can verify a signature, you must initiate the gadget with a curve configuration. 
+
+To initiate the curve:
 
 ```ts
 // create a secp256k1 curve
 class Secp256k1 extends createForeignCurve(Crypto.CurveParams.Secp256k1) {}
 ```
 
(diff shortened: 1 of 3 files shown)


docs/zkapps/o1js/ecdsa.mdx
```diff
@@ -22,48 +22,50 @@ Please note that zkApp programmability is not yet available on Mina Mainnet, but
...
 
-# Why ECDSA?
+## Why ECDSA?
 
-In order to interact with other blockchains or verify data from the outside world, o1js needs to be able to verify signatures.
-ECDSA is a widely used algorithm that is supported by many libraries and tools. It is also used in Ethereum, so it is important to support it in o1js as well. For example, Ethereum transactions are signed using ECDSA over the secp256k1 curve.
-If a zkApp developer wanted to verify an Ethereum transaction and make a statement about it, they would need to be able to verify the signature of the transaction. This is why ECDSA is important for zkApps.
+To interact with other blockchains and verify data from the outside world, o1js needs to be able to verify signatures.
+ECDSA is a widely used algorithm that is supported by many libraries and tools. For example, Ethereum transactions are signed using ECDSA over the secp256k1 curve.
+As a zkApp developer, when you want to verify an Ethereum transaction and make a statement about it, you must be able to verify the signature of the transaction which is why ECDSA is important for zkApps.
```
barriebyron, line 31: same style edits as Keccak, conforming to docs style guide and how we talk directly to our devs https://github.com/o1-labs/docs2/wiki/Docs-Style-Guide

docs/zkapps/o1js/ecdsa.mdx
```diff
@@ -22,48 +22,50 @@ Please note that zkApp programmability is not yet available on Mina Mainnet, but
...
 
-# Basic usage
+## Basic usage
 
-The ECDSA gadget is used to verify ECDSA signatures. It takes as input the message, the signature, and the public key of the signer. It outputs a `Bool` indicating whether the signature is valid or not.
+The ECDSA gadget is used to verify ECDSA signatures. The gadget takes as input the message, the signature, and the public key of the signer. It outputs a `Bool` indicating whether the signature is valid.
 
-Before we can verify a signature, we must initiate the gadget with a curve configuration. The most commonly used curve is secp256k1, which is also used in Ethereum. We can initiate the curve as follows:
+Before you can verify a signature, you must initiate the gadget with a curve configuration.
```
barriebyron, line 37: we already stated that sep256k1 is the most commonly used, so let's omit it here

docs/zkapps/o1js/ecdsa.mdx
```diff
@@ -22,48 +22,50 @@ Please note that zkApp programmability is not yet available on Mina Mainnet, but
...
 ```ts
 // create a secp256k1 curve
 class Secp256k1 extends createForeignCurve(Crypto.CurveParams.Secp256k1) {}
 ```
 
-By default, o1js exports a set of already defined curves. We can use the `createForeignCurve` function to create a curve from a `CurveParams` object. The `CurveParams` object contains the parameters of the curve, such as the modulus, the generator and the parameters `a` and `b` of the cruve equation `y^2 = x^3 + ax + b`.
+By default, o1js exports a set of predefined curves. You can use the `createForeignCurve` function to create a curve from a `CurveParams` object. The `CurveParams` object contains the parameters of the curve, such as the modulus, the generator, and the parameters `a` and `b` of the curve equation `y^2 = x^3 + ax + b`.
 
-The namespace `Crypto.CurveParams` exports a few pre-defined curves, such as `Pallas`, `Vesta` and `Secp256k1`.
+The namespace `Crypto.CurveParams` exports predefined curves, such as `Pallas`, `Vesta`, and `Secp256k1`.
```
barriebyron, line 48: @Trivo25  line 46 says that o1js exports a set of predefined curves (is that automagic buy default? or we need to include this code?

here we specify that it's the namespace that exports them... same? do we need both mentions?

barriebyron (approved):
excellent context and explanations for getting devs up to speed with ECDSA. 

Edits for docs style to conform with how we directly talk to our developers. A few questions for you @Trivo25

docs/zkapps/o1js/ecdsa.mdx
Trivo25: It's exported under the `Crypto` namespace and accessible as `Crypto.CurveParams` if thats what you mean

garwalsh (approved):

docs/zkapps/o1js/ecdsa.mdx
```diff
@@ -0,0 +1,128 @@
...
+let publicKey = Secp256k1.generator.scale(privateKey);
+
+// create a message, for a detailed explanation of `Bytes` take a look at the Keccak overview
+let message = Bytes32.fromString('cat');
+
+// sign a message - this is not a provable method!
+let signature = Ecdsa.sign(message.toBytes(), privateKey.toBigInt());
+```
+
+Most usages require you to verify a hash of a message instead of the message itself. In Ethereum, for example, the message is the [Keccak](keccak) hash of the transaction data.
```
mitschabaude, line 82: This is a bit misleading, because `sign` and `verify` already include Keccak256 hashing (yes, a bit opinionated). The text makes it sounds like the user might need to hash the message before verifying.

Lower-level methods which sign / verify the EC scalar (which is created from the hash), and thus allow to use any hash function, are available as `signHash()` and `verifySignedHash()`

Trivo25: I removed that part, I reference Keccak in a different section so that should be enough!

PR #790, opened 2024-01-06
Update docs/zkapps/tutorials/03-deploying-to-a-network

Violet-Bora-Lee:
- fix document errors


barriebyron (approved):
thanks for these updates @Violet-Bora-Lee !

PR #793, opened 2024-01-15
Fix: Remove security footgun example in onchain values docs

mitschabaude:
The code example deleted here (written by me originally) makes no sense and suggests usage of on-chain values in an insecure way: Using a completely unconstrained value (`now`) to define a range that treats that unconstrained value as the actual value.

A better example for `requireBetween()` with the `.get()` value as input would be to treat it like "`requireEquals()` but allowing for an error"

### DIFF
--- a/docs/zkapps/o1js/on-chain-values.mdx
+++ b/docs/zkapps/o1js/on-chain-values.mdx
@@ -64,13 +64,6 @@ class VotingApp extends SmartContract {
 
 A more refined example could store the current start date in an on-chain state variable, which can then be reset by some process that is also encoded by the zkApp.
 
-In addition to using a predefined range, you can also construct a range that depends on another variable, such as the current time. For example, a DEX with an order book could support orders that expire after an hour. In this case, the range would be `[now, now + 1h]`:
-
-```ts
-const now = this.network.timestamp.get();
-this.network.timestamp.requireBetween(now, now.add(60 * 60 * 1000));
-```
-
 ### Network reference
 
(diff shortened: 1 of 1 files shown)


docs/zkapps/o1js/on-chain-values.mdx
```diff
@@ -64,13 +64,6 @@ class VotingApp extends SmartContract {
 
 A more refined example could store the current start date in an on-chain state variable, which can then be reset by some process that is also encoded by the zkApp.
 
-In addition to using a predefined range, you can also construct a range that depends on another variable, such as the current time. For example, a DEX with an order book could support orders that expire after an hour. In this case, the range would be `[now, now + 1h]`:
```
mitschabaude, line 67: this example was never needed to convey that "you can also construct a range that depends on another variable". The sentence right above it, about storing the start date on the zkApp, is already an excellent example use case for constructing a range from a variable

barriebyron (approved):

barriebyron:
thanks for the fix!
@mitschabaude  would you consider the rest of this doc https://docs.minaprotocol.com/zkapps/o1js/on-chain-values to be technically accurate so we can close the HF Audit task? See https://app.zenhub.com/workspaces/o1js-docs-6565f4667e7d97084bc9895d/issues/gh/o1-labs/docs2/478

mitschabaude:
> @mitschabaude would you consider the rest of this doc https://docs.minaprotocol.com/zkapps/o1js/on-chain-values to be technically accurate so we can close the HF Audit task?

@barriebyron not sure, haven't gone through it all in detail again

PR #794, opened 2024-01-16
Fix square root example for o1js

satyambnsal:
Square root of `81` with o1js will not be `9`, it will be `28948022309329048855892746252171976963363056481941560715954676764349967630328`

### DIFF
--- a/docs/zkapps/o1js/basic-concepts.md
+++ b/docs/zkapps/o1js/basic-concepts.md
@@ -116,7 +116,7 @@ x = x.sub(1); // x = 6
 x = x.mul(3); // x = 18
 x = x.div(2); // x = 9
 x = x.square(); // x = 81
-x = x.sqrt(); // x = 9
+x = x.sqrt(); // x = 28948022309329048855892746252171976963363056481941560715954676764349967630328
 
 let b = x.equals(8); // b = Bool(false)
 b = x.greaterThan(8); // b = Bool(true)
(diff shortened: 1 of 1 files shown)


docs/zkapps/o1js/basic-concepts.md
```diff
@@ -116,7 +116,7 @@ x = x.sub(1); // x = 6
 x = x.mul(3); // x = 18
 x = x.div(2); // x = 9
 x = x.square(); // x = 81
-x = x.sqrt(); // x = 9
+x = x.sqrt(); // x = 28948022309329048855892746252171976963363056481941560715954676764349967630328
```
mitschabaude, line 119: nice catch :D let's change it so that the result makes more sense:

```suggestion
x = x.sqrt(); // x = -9
```

mitschabaude (approved):

barriebyron:
thanks for the fix @satyambnsal  and thanks for the committed suggestion @mitschabaude

docs/zkapps/o1js/basic-concepts.md
satyambnsal: Sorry, I didn't get it. is `28948022309329048855892746252171976963363056481941560715954676764349967630328` equivalent to `-9`

mitschabaude: yes! in modular arithmetic, -9 is p-9 where p is the field size. I checked that the square root equals that

PR #798, opened 2024-01-18
Correction of some part of codes in tutorial-07

kadirchan:
The parts in the tutorial did not match each other in some places. I corrected these parts as done in https://github.com/o1-labs/docs2/tree/main/examples/zkapps/07-oracles.


barriebyron (approved):
Thank you for this valuable fix @kadirchan !

PR #799, opened 2024-01-18
Improvements on tutorial-08

kadirchan:
I think it will be better like that.


barriebyron (approved):
We have a variety of opinions on whether we need the ramp up steps for later tutorials, so this tutorial 8 was published without them. Assuming devs would start with tutorial 1 and go forward, the basic project setup knowledge is assumed. Those are two assumptions! Now that you have contributed this improvement, I see no reason not to provide more context to help our devs succeed. Thank you for this update!

kadirchan:
Yes, you are right, these steps are done many times in the previous sections, but I thought it would be good for developers starting from the middle of the book to benefit from this. You are welcome @barriebyron !

PR #820, opened 2024-01-31
Feature: Adding new block explorer - minataur

kogeletey:
Unlike the previous two, its code is open, and it works super fast.


barriebyron:
Thanks for this PR @kogeletey 

We have only a single mention of Minataur so far, and it's in the glossary https://docs.minaprotocol.com/glossary#block-explorer 

To be complete, would you be up for adding Minataur to the FAQ?
https://github.com/o1-labs/docs2/blob/main/docs/about-mina/faq.mdx#L51

We will make other related doc updates regarding block explorers before the major upgrade

kogeletey:
It seems like I added it to the [faq](https://github.com/kogeletey/mina-docs/blob/feat/minataur-block-explorer/docs/about-mina/faq.mdx) and even to [how-to-send-and-receive.mdx](https://github.com/o1-labs/docs2/blob/cf1f4cd035a7e6c2390d86363f954b2292528ec7/docs/using-mina/how-to-send-and-receive.mdx) files. 
What else should I do?

barriebyron (approved):
thanks for this update @kogeletey

barriebyron (approved):

PR #822, opened 2024-02-01
fix: minor doc fixes for style, consistency

barriebyron:
This PR addresses minor style updates and does not contain substantive updates


docs/zkapps/o1js/actions-and-reducer.mdx
```diff
@@ -44,7 +44,7 @@ On a `reducer`, you have two functions: `reducer.dispatch()` and `reducer.reduce
 this.reducer.dispatch(Field(1000));
 ```
 
-"Reduce" is more involved, but it gives you full power to process actions however it suits your application. It might be easiest to grasp from an example. Say we have a list of actions and want to find out if one of them is equal to `1000`. In JavaScript, there's a built-in function on `Array` which does this:
+"Reduce" is more involved, but it gives you full power to process actions however it suits your application. It might be easiest to grasp from an example. Say you have a list of actions and want to find out if one of them is equal to `1000`. In JavaScript, there's a built-in function on `Array` which does this:
```
iregina, line 47: @barriebyron Just curious - are we using "you" for docs?

barriebyron: very good question, and the answer is YES
We write to the devs and for the devs
https://github.com/o1-labs/docs2/wiki/Docs-Style-Guide#language-and-grammar

iregina (approved):
Approve

PR #833, opened 2024-02-06
Lightnet updates for deploy workflow 

barriebyron:
DX/UX focus on deploying zkApps 
differences between local simulated blockchain and Mainnet/Berkeley


barriebyron:
duplicate PR, fixed in https://github.com/o1-labs/docs2/pull/869

PR #846, opened 2024-02-13
Add a new chapter for common types and expand the basic concepts

yunus433:
The first commit to update o1js docs based on the O(1) Labs feedback. Changed common types chapter and expanded basic concepts. Added the following information:

- Detailed Field description
- Explanation on assertions
- Information on printing
- Detailed Bool description
- Added UInt32 / UInt64
- Added Public / Private Keys
- Expanded conditions
- Added loops
- Expanded functions

### DIFF
--- a/docs/zkapps/o1js/basic-concepts.md
+++ b/docs/zkapps/o1js/basic-concepts.md
@@ -29,11 +29,11 @@ zkApp programmability is not yet available on the Mina Mainnet. You can get star
 
 o1js, fka. SnarkyJS, is a TypeScript (TS) library for writing general-purpose ZK programs and writing ZK smart contracts for the Mina Blockchain. In order to create a ZKP (zero knowledge proof), you should use types and operations that can be converted into a ZKP. o1js gives you a lot of different built in types and customizable structures that you can use to create ZKPs.
 
-## Field Type
+## Field
 
-`Field` elements are the basic unit of data in ZK programming. Every other data type (either built-in or composite) in o1js is made up of `Field` elements.
+Field elements are the basic unit of data in ZK programming. Every other data type (either built-in or composite) in o1js is made up of field elements.
 
-Each `Field` element can store a number up to almost 256 bits in size. You can think of it as a `uint256` in Solidity, but there are some fundamental differences explained below.
+Each `Field` element in o1js can store a number up to almost 256 bits in size. You can think of it as a `uint256` in Solidity, but there are some fundamental differences explained below.
 
 :::note
 
@@ -41,62 +41,61 @@ For the cryptography inclined, the prime order of the o1js `Field` is: 28,948,02
 
 :::
 
-## Creating a Field Element
+## Creating a `Field` Element
 
-You can create a `Field` either with the `new` constructor or by directly calling the `Field` as a constructor. The best practice is to use `Field` without the `new` constructor.
+You can create a `Field` either with the `new` constructor or by directly calling the `Field` as a constructor. The convention is to use `Field` without the `new` constructor.
 
 <!-- prettier-ignore -->
 ```ts
-const x = new Field(42);
-const y = Field(923); // Best practice
+let x = Field(42);
 ```
 
 You can create a `Field` from anything that is "field-like": `number`, `string`, `bigint`, or another `Field`. Note that the argument you give must be an integer in the `Field` range.
 
 <!-- prettier-ignore -->
 ```ts
-const x = Field("12347"); // From string
-const y = Field(172384782343434n); // From bigint
-const z = Field(x); // From another Field
+let x = Field("12347"); // From string
+let y = Field(172384782343434n); // From bigint
+let z = Field(x); // From another Field
 
-const k = Field("asdf"); // Throws, as this is not a number
-const l = Field("234.34"); // Throws, as this is not an integer
+let k = Field("asdf"); // Throws, as this is not a number
+let l = Field("234.34"); // Throws, as this is not an integer
 ```
 
-Nevertheless, you can create a `Field` from a negative number. This method implicity calculates the [modular inverse](https://en.wikipedia.org/wiki/Modular_multiplicative_inverse) of the given argument. This is an advanced level information, and you do not need it for simple zkApps.
+Nevertheless, you can create a `Field` from a negative number. This method implicity calculates the additive modular inverse of the given argument. This is an advanced level information, and you do not need it for simple zkApps.
 
 <!-- prettier-ignore -->
 ```ts
-const x = Field(-1); // If you try printing this to the screen, you will see a very big integer
+let x = Field(-1); // If you try printing this to the screen, you will see a very big integer
 ```
 
-## Methods in the Field Type
+## Methods on the `Field` Type
 
 `Field` type includes a lot of built in methods. You can call all methods in the `Field` type either with a `Field` or with a "field-like" value. In the latter case, the argument is implicity converted to a `Field`.
 
-All common arithmetic operations exist in the `Field` type. They create and return a new `Field` element.
+All common arithmetic operations are available on the `Field` type. They create and return a new `Field` element.
 
 <!-- prettier-ignore -->
 ```ts
-const x = Field(45);
-const y = Field(78);
+let x = Field(45);
+let y = Field(78);
 
-const sum = x.add(y);
-const sub = x.sub(y);
-const mul = x.mul(47); // You can also use "field-like" values
-const div = x.div(89); // This is a modular division
-const neg = x.neg(); // This is equivalent to x.mul(-1)
+let sum = x.add(y);
+let sub = x.sub(y);
+let mul = x.mul(47); // You can also use "field-like" values
+let div = x.div(89); // This is a modular division
+let neg = x.neg(); // This is equivalent to x.mul(-1)
 
-const square = x.square(); // x^2, equivalent to x.mul(x)
-const sqrt = x.sqrt(); // Modular square root of x
-const inv = x.inv(); // Modular inverse of x
+let square = x.square(); // x^2, equivalent to x.mul(x)
+let sqrt = x.sqrt(); // Modular square root of x
+let inv = x.inv(); // Modular inverse of x
 ```
 
 You can also perform logical operations and comparisons with `Field` elements. These methods return a [Bool](./common-types-and-functions.md#bool) element, which is another common o1js built in type, equivalent to `boolean` in other languages and frameworks.
 
 <!-- prettier-ignore -->
 ```ts
-const x = Field(45);
+let x = Field(45);
 
 x.equals(45); // True
 x.greaterThan(45); // False
@@ -114,7 +113,7 @@ If you want to reject a TX (transaction) in your zkApp smart contract, you can u
 
 <!-- prettier-ignore -->
 ```ts
-const x = Field(45);
+let x = Field(45);
 
 x.assertEquals(45); // True - passes
 x.assertGreaterThan(45); // False - throws
@@ -127,25 +126,106 @@ You can think assertions in o1js similar to `require()` function in Solidity.
 
 ## Printing a Field to the Screen
 
-`Field` elements are stored as classes in o1js. This is why if you print a `Field` without a conversion, you will get an unreadable output. Hopefully, o1js gives a variety of conversion methods to print `Field` elements to the screen.
+`Field` elements are stored as classes in o1js. This is why if you print a `Field` using `console.log`, you will get an unreadable output. Hopefully, o1js provides an equivalent function to `console.log` for provable types: `Provable.log`
 
 <!-- prettier-ignore -->
 ```ts
-const x = Field(45);
-
-console.log(x.toString()); // string
-console.log(x.toBigInt()); // bigint
-console.log(x.toJSON()); // JSON string
-```
-
-However, note that these methods are _not_ provable, meaning you cannot use them in your final zkApp as a part of the ZKP. Use them only to debug your code.
-
-<!-- prettier-ignore -->
-```ts
-const x = Field(45);
-const y = Field(x.toString()); // You should NOT write this
+let x = Field(45);
+Provable.log(x); // Prints 45
 ```
 
 ## Other Types and Methods
 
 In addition to the `Field` type, o1js gives you a variety structures to help you build a powerful zkApp. You can learn more about them in the next chapter, [Common Types and Functions](./common-types-and-functions.md).
+
+:::note
+
+The information given until here is more than enough for you to start creating powerful zkApps with o1js. For more advanced level content, you can read the following sections. If you are new to o1js, just continue with the next chapter, [Common Types and Functions](./common-types-and-functions.md).
+
+:::
+
+## Deeper into the Field Type
+
+`Field` element is a positive integer in a prime order [finite field](https://en.wikipedia.org/wiki/Finite_field), which is a set of integers in between 1 and the prime order of the field (a.k.a. an arbitrary prime number describing the field).
+
+The order of the o1js `Field` can be accessed as a static property of the class.
+
+<!-- prettier-ignore -->
+```ts
+console.log(Field.ORDER);
+```
+
+When you perform an operation inside a finite field, the result always stays in the range. This is achieved through [modular arithmetic](https://en.wikipedia.org/wiki/Modular_arithmetic). This behaviour may be unfamiliar to you if you have not worked with ZKPs before. Let us explain more with some examples.
+
+In the addition or multiplication operations where the result is bigger than the order of the finite field, you receive the modulus of the result with the prime order of the field.
+
+<!-- prettier-ignore -->
+```ts
+Provable.log(Field(Field.ORDER - 1) + Field(2)); // This will print 1
+```
+
+If you try creating a `Field` from a negative number you will receive the modular additive inverse of the number. For instance `(1 + (Field.ORDER - 1)) % Field.ORDER` is 0, so the additive inverse of 1 is `Field.ORDER - 1` in this field.
+
+<!-- prettier-ignore -->
+```ts
+Provable.log(Field(-1)); // This will print `Field.ORDER - 1`
+```
+
+Similarly, if your substraction results in a negative number you will get the modular additive inverse.
+
+For division, if the result is not an integer, you receive the [modular multiplicative inverse](https://en.wikipedia.org/wiki/Modular_multiplicative_inverse) of the division: If `(x * y) % p` is `z` (where `p` is the order of the finite field), then the modular multiplicative inverse of the division `z / x` equals `y`. Note that `x`, `y`, and `z` are all integers.
+
+<!-- prettier-ignore -->
+```ts
+Provable.log(Field(3).div(Field(5))); // This will print a very big integer
+```
+
+Because of this behaviour, it is usually not very useful to use the final result of an uneven division. Instead, you can prove the result with multiplication.
+
+<!-- prettier-ignore -->
+```ts
+// We want x, an argument of the smart contract, to be equal to 7 / 3
+x.mul(Field(3)).assertEquals(Field(7));
+```
+
+If you want to have integer division instead, you can use `UInt32` or `UInt64` types (see [Common Types and Functions](./common-types-and-functions.md)).
+
+## Provability and Provable Code in o1js
+
+We have stated many times above that all the code you write in o1js must be provable, but what is provability?
+
+A provable code can be thought as a code that o1js compiler can convert into a ZKP. Just because you write something with o1js, it does not become automatically provable. You must follow all the rules above and always use provable types in your code.
+
+However, provable types are created from regular TS variables.
+
+<!-- prettier-ignore -->
+```ts
+const x = 42; // This is TS number
+let x_provable = Field(x); // This code is perfectly valid
+```
+
+In reality, the `x_provable` variable in the example above is a _constant_ `Field` variable. By constant, we do not mean that it is a TS constant (which you define by using `const`), but we say that the value of `x_provable` is known at the compilation time. When you receive an argument to your smart contract method, then the argument is not a constant, thus you should only use provable methods on it.
+
+<!-- prettier-ignore -->
+```ts
+@method foo(x: Field) {
+  // Here, the value of x depends on the user input, thus type of x must be provable
+}
+```
+
+You can learn if a variable is a constant or not in o1js by using `isConstant()` method. Note that this is a function mostly for internal use, but it might give you a good insight on how to work with types in o1js.
+
+<!-- prettier-ignore -->
+```ts
+let x = Field(42);
+console.log(x.isConstant()); // This will print true
+```
+
+Always remember that o1js is used for writing ZKPs, and what you write with o1js must have a sense in the final proof. The public or private inputs of a ZKP may change depending on the user input, but once the inputs are provided, the final ZKP must make sure they are not changed. By using provable o1js variables, you make sure that your variables are always consistent with your proof logic.
+
+<!-- prettier-ignore -->
+```ts
+@method foo(x: Field, y: Field) {
+  x.add(y).assertEquals(5); // we know neither the value of x, nor of the y, but we are sure that (x + y) % Field.ORDER equals 5
+}
+```
--- a/docs/zkapps/o1js/common-types-and-functions.md
+++ b/docs/zkapps/o1js/common-types-and-functions.md
@@ -39,31 +39,33 @@ o1js includes built in types and methods that you can use inside your zkApp and 
 
 <!-- prettier-ignore -->
 ```ts
-const x = new Bool(true);
-const y = new Bool(false);
+let x = new Bool(true);
+let y = new Bool(false);
 ```
 
-With the `Bool` type, you can call the contructor without `new`, just like the `Field` type. The best practice is to use the constructor function without `new`:
+With the `Bool` type, you can call the contructor without `new`, just like the `Field` type. The convention is to use the constructor function without `new`:
 
 <!-- prettier-ignore -->
 ```ts
-const x = Bool(true);
-const y = Bool(false);
+let x = Bool(true);
+let y = Bool(false);
 ```
 
 You can also create a `Bool` from another `Bool`.
 
 <!-- prettier-ignore -->
 ```ts
-const x = Bool(true);
-const y = Bool(x);
+let x = Bool(true);
+let y = Bool(x);
 ```
 
 It is also possible to convert a `Field` into a `Bool`. However, you should make sure that the `Field` is "bool-like" (with a value either 1 or 0) before converting it.
 
 ```ts
 let x = Field(1);
-x.assertBool(); // Asserts that x has a value either 1 or 0
+
+// The line below makes sure x can be safely converted into a `Bool`.
+x.equals(Field(1)).or(x.equals(Field(0))).assertTrue(); // You will learn about the methods used here below.
 
 let x_bool = Bool(x.value);
 ```
@@ -72,25 +74,25 @@ let x_bool = Bool(x.value);
 
 <!-- prettier-ignore -->
 ```ts
-const x = Bool(true);
-const y = Bool(false);
-const z = y.not(); // True
+let x = Bool(true);
+let y = Bool(false);
+let z = y.not(); // True
 
-const equals = x.equals(y); // False
-const and = x.and(y); // False
-const or = x.or(y); // True
+let equals = x.equals(y); // False
+let and = x.and(y); // False
+let or = x.or(y); // True
 
-const equals_from_boolean = x.equals(true); // True
-const and_from_boolean = x.and(true); // True
-const or_from_boolean = y.or(false); // False
+let equals_from_boolean = x.equals(true); // True
+let and_from_boolean = x.and(true); // True
+let or_from_boolean = y.or(false); // False
 ```
 
-You can also assert `Bool` variables. When you assert something, your zkApp method throws if the assertion fails.
+You can also assert `Bool` variables. When you assert something, your zkApp method throws if the assertion fails. You can read more about assertions in the [basic concepts](./basic-concepts.md#assertions-in-o1js) chapter.
 
 <!-- prettier-ignore -->
 ```ts
-const x = Bool(true);
-const y = Bool(false);
+let x = Bool(true);
+let y = Bool(false);
 
 x.assertEquals(y); // Throws
 x.assertTrue(); // Nothing happens, as assertion passes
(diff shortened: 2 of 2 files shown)


docs/zkapps/o1js/basic-concepts.md
```diff
@@ -27,109 +27,125 @@ zkApp programmability is not yet available on the Mina Mainnet. You can get star
...
 
-Field elements are the basic unit of data in zero-knowledge proof programming. Each field element can store a number up to almost 256 bits in size. You can think of it as a uint256 in Solidity.
+`Field` elements are the basic unit of data in ZK programming. Every other data type (either built-in or composite) in o1js is made up of `Field` elements.
+
+Each `Field` element can store a number up to almost 256 bits in size. You can think of it as a `uint256` in Solidity, but there are some fundamental differences explained below.
 
 :::note
 
-For the cryptography inclined, the exact max value that a field can store is: 28,948,022,309,329,048,855,892,746,252,171,976,963,363,056,481,941,560,715,954,676,764,349,967,630,336
+For the cryptography inclined, the prime order of the o1js `Field` is: 28,948,022,309,329,048,855,892,746,252,171,976,963,363,056,481,941,560,715,954,676,764,349,967,630,336
```
mitschabaude, line 40: This is an improvement to before. But what would also be interesting for cryptographically inclined readers is to know what kind of field this is (Pallas base field) with a link to more details.

Also, I suggest to write the modulus in hex format, because that visually shows the twoadicity

docs/zkapps/o1js/basic-concepts.md
```diff
@@ -27,109 +27,125 @@ zkApp programmability is not yet available on the Mina Mainnet. You can get star
...
 
 :::
 
-For example, in typical programming, you might use:
-
-`const sum = 1 + 3`.
+## Creating a Field Element
 
-In o1js, you write this as:
+You can create a `Field` either with the `new` constructor or by directly calling the `Field` as a constructor. The best practice is to use `Field` without the `new` constructor.
```
mitschabaude, line 46: Suggestion: "common practice" or "convention" instead of "best practice"

Also, I wouldn't even show the version with new in the code example

docs/zkapps/o1js/basic-concepts.md
```diff
@@ -27,109 +27,125 @@ zkApp programmability is not yet available on the Mina Mainnet. You can get star
...
+const y = Field(172384782343434n); // From bigint
+const z = Field(x); // From another Field
 
-## Built-in data types
+const k = Field("asdf"); // Throws, as this is not a number
+const l = Field("234.34"); // Throws, as this is not an integer
+```
 
-Some common data types you may use are:
+Nevertheless, you can create a `Field` from a negative number. This method implicity calculates the [modular inverse](https://en.wikipedia.org/wiki/Modular_multiplicative_inverse) of the given argument. This is an advanced level information, and you do not need it for simple zkApps.
```
mitschabaude, line 66: This is incorrect. A negative input causes it to create the additive inverse in the field, i.e. takes the number modulo the field size. But "modular inverse" refers to the _multiplicative inverse_ and you also link to an article about multiplicative inverses

docs/zkapps/o1js/basic-concepts.md
```diff
@@ -27,109 +27,125 @@ zkApp programmability is not yet available on the Mina Mainnet. You can get star
...
-new Field(x);  // accepts an integer, or a numeric string if you want to represent a number greater than JavaScript can represent but within the max value that a field can store.
-new UInt64(x); // accepts a Field - useful for constraining numbers to 64 bits
-new UInt32(x); // accepts a Field - useful for constraining numbers to 32 bits
+const x = Field(-1); // If you try printing this to the screen, you will see a very big integer
+```
 
-PrivateKey, PublicKey, Signature; // useful for accounts and signing
-new Group(x, y); // a point on our elliptic curve, accepts two Fields/numbers/strings
-Scalar; // the corresponding scalar field (different than Field)
+## Methods in the Field Type
```
mitschabaude, line 73:
```suggestion
## Methods on the `Field` type
```

docs/zkapps/o1js/basic-concepts.md
```diff
@@ -27,109 +27,125 @@ zkApp programmability is not yet available on the Mina Mainnet. You can get star
...
 
-CircuitString.from('some string'); // string of max length 128
-```
+`Field` type includes a lot of built in methods. You can call all methods in the `Field` type either with a `Field` or with a "field-like" value. In the latter case, the argument is implicity converted to a `Field`.
 
-In the case of `Field` and `Bool`, you can also call the constructor without `new`:
+All common arithmetic operations exist in the `Field` type. They create and return a new `Field` element.
```
mitschabaude, line 77:
```suggestion
All common arithmetic operations are available on the `Field` type. They create and return a new `Field` element.
```

docs/zkapps/o1js/basic-concepts.md
```diff
@@ -27,109 +27,125 @@ zkApp programmability is not yet available on the Mina Mainnet. You can get star
...
+x.assertGreaterThanOrEquals(45); // True - passes
+x.assertLessThan(45); // False - throws
+x.assertLessThanOrEquals(45); // True - passes
 ```
 
-## Functions
+You can think assertions in o1js similar to `require()` function in Solidity.
 
-Functions work as you would expect in TypeScript. For example:
+## Printing a Field to the Screen
```
mitschabaude, line 128: I'm not a fan of this section at all, because toString etc don't work inside a circuit. Why not teach about Provable.log() here?

mitschabaude: Also teaching about constants vs variables would be nice

docs/zkapps/o1js/common-types-and-functions.md
```diff
@@ -0,0 +1,354 @@
...
+const y = Bool(x);
+```
+
+It is also possible to convert a `Field` into a `Bool`. However, you should make sure that the `Field` is "bool-like" (with a value either 1 or 0) before converting it.
+
+```ts
+let x = Field(1);
+x.assertBool(); // Asserts that x has a value either 1 or 0
+
+let x_bool = Bool(x.value);
```
mitschabaude, line 68: This needs to be replaced with `Bool.Unsafe.ofField`. it's important that this API is _explicit_ about its unsafety. The API you show is internal and it's even problematic that it type checks

docs/zkapps/o1js/common-types-and-functions.md
```diff
@@ -0,0 +1,354 @@
...
+<!-- prettier-ignore -->
+```ts
+const x = Bool(true);
+const y = Bool(false);
+
+x.assertEquals(y); // Throws
+x.assertTrue(); // Nothing happens, as assertion passes
+x.assertFalse(); // Throws
+```
+
```
mitschabaude, line 99: Would be nice to teach about assert() here

docs/zkapps/o1js/common-types-and-functions.md
```diff
@@ -0,0 +1,354 @@
...
+<!-- prettier-ignore -->
+```ts
+const x = Bool(true);
+const y = Bool(false);
+
+x.not().and(y).or(true); // True
+```
+
+**Important:** All logical expressions are executed in the method call order, as with any other regular language or framework. However, as these are logical statements inside ZKPs, all expressions are always evaluated regardless of the logical ordering. Please see conditionals chapter [below](./common-types-and-functions.md#conditionals) for more details.
```
mitschabaude, line 110: Nice call-out!

docs/zkapps/o1js/common-types-and-functions.md
```diff
@@ -0,0 +1,354 @@
...
+
+## UInt32
+
+`UInt32` type is a 32 bit positive integer. You can create it using the `new` constructor or the `UInt32.from()` static method. The best practice is to use `UInt32.from()`.
+
+Both the constructor and `UInt32.from()` method accepts `number`, `string`, `bigint`, `Field`, and `UInt32` type. In order the conversion to be successful, the type should evaluate to "uint32-like", meaning an integer in the range [0, $2^{32}$].
```
mitschabaude, line 116: The range is incorrect, 2^32 - 1 is the max

docs/zkapps/o1js/common-types-and-functions.md
```diff
@@ -0,0 +1,354 @@
...
+const y = x.value; // This is a Field with the value 45
+```
+
+## Public and Private Keys
+
+For accessing Mina blockchain accounts, you can use `PublicKey` and `PrivateKey` types of o1js. You can create a key from a [base58 string](https://en.wikipedia.org/wiki/Binary-to-text_encoding).
+
+<!-- prettier-ignore -->
+```ts
+const x_private = PrivateKey.fromBase58("base58_private_key");
```
mitschabaude, line 232: It's weird to not show a real private key here

docs/zkapps/o1js/common-types-and-functions.md
```diff
@@ -0,0 +1,354 @@
...
+const x = PublicKey.fromBase58("B62qq8sm7JdsED6VuDKNWKLAi1Tvz1jrnffuud5gXMq3mgtd");
+const y = PublicKey.fromJSON("B62qq8sm7JdsED6VuDKNWKLAi1Tvz1jrnffuud5gXMq3mgtd");
+```
+
+You can also create a random `PrivateKey`, an empty `PublicKey`, or convert a `PrivateKey` to a `PublicKey`.
+
+<!-- prettier-ignore -->
+```ts
+const random = PrivateKey.random();
+const empty = PublicKey.empty();
```
mitschabaude, line 244: If you're not going to explain what an empty public key is, leave it out

docs/zkapps/o1js/common-types-and-functions.md
```diff
@@ -0,0 +1,354 @@
...
+```ts
+const x = Provable.if(
+  Bool(foo),
+  (() => { // Here, we use arrow functions to return a Field element, but you can use any function that returns a provable type
+    return Field(5).div(3);
+  })(),
+  (() => {
+    return Field(5).div(2);
+  })()
+); // This variable is of type Field
```
mitschabaude, line 286: I dislike this way of presenting provable.if, because it for newbs it creates the impression that code is executed conditionally (I've experienced this)

Also, it's not cleaner than computing the two outcomes before selecting them with if. I recommend to show an example for that instead

docs/zkapps/o1js/common-types-and-functions.md
```diff
@@ -0,0 +1,354 @@
...
+
+<!-- prettier-ignore -->
+```ts
+// This will NOT work
+while (foo) {
+  x.add(y);
+}
+```
+
+You can overcome this limitation by using static sized loops that iterate always the same amount of time. You can perform provable operations with these loops using the `Provable.if`.
```
mitschabaude, line 301:
```suggestion
You can overcome this limitation by using static sized loops that iterate always the same amount of time. You can perform provable operations with these loops using `Provable.if()`.
```

docs/zkapps/o1js/common-types-and-functions.md
```diff
@@ -0,0 +1,354 @@
...
+
+<!-- prettier-ignore -->
+```ts
+function powerOfTwo(pow: Field): Field {
```
mitschabaude, line 305: This example is super inefficient due to its reliance on expensive field comparisons. It needs to be improved

docs/zkapps/o1js/common-types-and-functions.md
```diff
@@ -0,0 +1,354 @@
...
+  const MAX_ITERATION_COUNT = 64;
+
+  pow.assertGreaterThanOrEqual(0);
```
mitschabaude, line 308: This is always true. Remove

docs/zkapps/o1js/common-types-and-functions.md
```diff
@@ -0,0 +1,354 @@
...
+  pow.assertSmallerThanOrEqual(MAX_ITERATION_COUNT); // This code is to calculate up to 2^64, you can optimize this based on your needs
```
mitschabaude, line 309: Use a 64 bit range check, or pass in a UInt64

docs/zkapps/o1js/common-types-and-functions.md
```diff
@@ -0,0 +1,354 @@
...
+
+  let result = Field(1); // Result of the calculation
+  let power = Field(0); // Current power of the iteration
+  let loopEnded = Bool(false); // Checking if the loop has ended yet
+
+  for (let i = 0; i <= MAX_ITERATION_COUNT; i++) { // Static sized loop
+    loopEnded = Provable.if( // End Condition
```
mitschabaude, line 316: Redundant if

docs/zkapps/o1js/common-types-and-functions.md
```diff
@@ -0,0 +1,354 @@
...
+      Field(i).greaterThanOrEqual(power),
```
mitschabaude, line 317:
```suggestion
      power.equals(i),
```

docs/zkapps/o1js/common-types-and-functions.md
```diff
@@ -0,0 +1,354 @@
...
+    ));
+  };
+
+  return result;
+};
+```
+
+# Functions
+
+Functions work as you would expect in TS. However, your functions inside ZKPs must take provable arguments and return provable types, just like anything in o1js. For example:
```
mitschabaude, line 339: That's not correct. It's common for functions to take parameters that define the circuit, not only variables.

docs/zkapps/o1js/common-types-and-functions.md
```diff
@@ -0,0 +1,354 @@
...
+
+<!-- prettier-ignore -->
+```ts
+function addOneAndDouble(x: Field): Field {
+  return x.add(1).mul(2);
+}
+```
+
+**Warning:** You should _not_ use recursion with functions in o1js. If you want to recurse a ZKP, you can use [o1js recursion](./recursion.mdx).
```
mitschabaude, line 348: Incorrect, if the recursion terminates on a condition that's static

mitschabaude (commented):
First pass from me

docs/zkapps/o1js/common-types-and-functions.md
```diff
@@ -0,0 +1,354 @@
...
+`Provable.if()` allows you to include any logical computation you need inside your ZKP. There is nothing wrong with evaluating more complex expressions inside `Provable.if()` and returning the final result.
+
+However, it works differently from ternary statements in the other languages and frameworks. Both arguments of `Provable.if()` _always_ executes, regardless of the truth value of the condition. `Provable.if()` choses which one to return based on the truth value and sets the new variable. This is why you should never assign anything outside of the if scope or assert anything.
+
+<!-- prettier-ignore -->
+```ts
+const x = Provable.if(
+  Bool(foo),
+  (() => { // Here, we use arrow functions to return a Field element, but you can use any function that returns a provable type
+    return Field(5).div(3);
```
mitschabaude, line 281: I'm not a big fan of how all our examples only ever use constant field elements as examples, where none of these circuit-writing rules and caveats actually apply

Explaining constants vs variables and then using variables in examples seems an important step in moving to more in-depth docs about circuit writing

docs/zkapps/o1js/common-types-and-functions.md
barriebyron, line 110:
```suggestion
**Important:** All logical expressions are executed in the method call order, as with any other regular language or framework. However, as these are logical statements inside ZKPs, all expressions are always evaluated regardless of the logical ordering. See [Conditionals](./common-types-and-functions.md#conditionals) for more details.
```

barriebyron: I agree with Gregor, nice work @yunus433 !
Guidance is to omit "please", see [Politeness and use of please](https://developers.google.com/style/tone#politeness-and-use-of-please)

Instead of "below" or "here" be sure to use short [descriptive link text](https://developers.google.com/style/link-text) that helps provide context for the material that you're linking to.

docs/zkapps/o1js/basic-concepts.md
```diff
@@ -27,109 +27,125 @@ zkApp programmability is not yet available on the Mina Mainnet. You can get star
...
 # o1js Basic Concepts
 
-o1js, fka. SnarkyJS, is a TypeScript (TS) library for writing general-purpose zk programs and writing zk smart contracts for Mina.
+o1js, fka. SnarkyJS, is a TypeScript (TS) library for writing general-purpose ZK programs and writing ZK smart contracts for the Mina Blockchain. In order to create a ZKP (zero knowledge proof), you should use types and operations that can be converted into a ZKP. o1js gives you a lot of different built in types and customizable structures that you can use to create ZKPs.
 
-## Field
+## Field Type
 
-Field elements are the basic unit of data in zero-knowledge proof programming. Each field element can store a number up to almost 256 bits in size. You can think of it as a uint256 in Solidity.
+`Field` elements are the basic unit of data in ZK programming. Every other data type (either built-in or composite) in o1js is made up of `Field` elements.
```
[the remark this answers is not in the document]
barriebyron: thank you @mitschabaude  I updated our Word list with this guidance
https://github.com/o1-labs/docs2/wiki/Word-List#field-element

docs/zkapps/o1js/common-types-and-functions.md
```diff
@@ -0,0 +1,354 @@
...
+const x = Provable.if(
+  Bool(foo), // The condition to evaluate inside if, must be of type Bool
+  a, // You can return any provable type from Provable.if
+  b // a and b must be of the same provable type
+); // Behaves like `foo ? a : b`
+```
+
+`Provable.if()` allows you to include any logical computation you need inside your ZKP. There is nothing wrong with evaluating more complex expressions inside `Provable.if()` and returning the final result.
+
+However, it works differently from ternary statements in the other languages and frameworks. Both arguments of `Provable.if()` _always_ executes, regardless of the truth value of the condition. `Provable.if()` choses which one to return based on the truth value and sets the new variable. This is why you should never assign anything outside of the if scope or assert anything.
```
barriebyron, line 274:
```suggestion
However, it works differently from ternary statements in the other languages and frameworks. Both arguments of `Provable.if()` _always_ execute, regardless of the truth value of the condition. `Provable.if()` chooses which one to return based on the truth value and sets the new variable. This is why you should never assign anything outside of the scope or assert anything.
```

docs/zkapps/o1js/common-types-and-functions.md
yunus433: Can I directly use `x.equals(Field(1)).or(x.equals(Field(0))).assertTrue()`, maybe that would be better than giving the internal API as an example? Or else I can directly delete this section, but I think it is important to show how conversions should work in o1js to new comers. Thank you @mitschabaude

docs/zkapps/o1js/basic-concepts.md
```diff
@@ -27,109 +27,205 @@ zkApp programmability is not yet available on the Mina Mainnet. You can get star
 
 # o1js Basic Concepts
 
-o1js, fka. SnarkyJS, is a TypeScript (TS) library for writing general-purpose zk programs and writing zk smart contracts for Mina.
+o1js, fka. SnarkyJS, is a TypeScript (TS) library for writing general-purpose ZK programs and writing ZK smart contracts for the Mina Blockchain. In order to create a ZKP (zero knowledge proof), you should use types and operations that can be converted into a ZKP. o1js gives you a lot of different built in types and customizable structures that you can use to create ZKPs.
```
barriebyron, line 30:
```suggestion
o1js, fka. SnarkyJS, is a TypeScript (TS) library for writing general-purpose ZK programs and writing ZK smart contracts for the Mina Blockchain. To create a ZKP (zero knowledge proof), you use types and operations that can be converted into a ZKP. o1js provides different built-in types and customizable structures that you can use to create ZKPs.
```

docs/zkapps/o1js/common-types-and-functions.md
mitschabaude: I think it's useful to show the conversion. Just use
```ts
let xBool = Bool.Unsafe.ofField(x);
```

docs/zkapps/o1js/basic-concepts.md
```diff
@@ -29,74 +29,73 @@ zkApp programmability is not yet available on the Mina Mainnet. You can get star
...
+You can create a `Field` either with the `new` constructor or by directly calling the `Field` as a constructor. The convention is to use `Field` without the `new` constructor.
 
 <!-- prettier-ignore -->
 ```ts
-const x = new Field(42);
-const y = Field(923); // Best practice
+let x = Field(42);
 ```
 
 You can create a `Field` from anything that is "field-like": `number`, `string`, `bigint`, or another `Field`. Note that the argument you give must be an integer in the `Field` range.
```
mitschabaude, line 53: >  Note that the argument you give must be an integer in the `Field` range.

Can you remove the second sentence here? It's not true, an integer outside the field range is also accepted and reduced modulo the field size (as you show below, with negative inputs)

mitschabaude: Or just "must be an integer"

docs/zkapps/o1js/basic-concepts.md
```diff
@@ -127,25 +126,106 @@ You can think assertions in o1js similar to `require()` function in Solidity.
...
+
+:::note
+
+The information given until here is more than enough for you to start creating powerful zkApps with o1js. For more advanced level content, you can read the following sections. If you are new to o1js, just continue with the next chapter, [Common Types and Functions](./common-types-and-functions.md).
+
+:::
+
+## Deeper into the Field Type
+
+`Field` element is a positive integer in a prime order [finite field](https://en.wikipedia.org/wiki/Finite_field), which is a set of integers in between 1 and the prime order of the field (a.k.a. an arbitrary prime number describing the field).
```
mitschabaude, line 149: Actually it's the set of integers between 0 and p-1, where p is the prime order!

docs/zkapps/o1js/basic-concepts.md
```diff
@@ -127,25 +126,106 @@ You can think assertions in o1js similar to `require()` function in Solidity.
...
-console.log(x.toJSON()); // JSON string
+<!-- prettier-ignore -->
+```ts
+console.log(Field.ORDER);
 ```
 
-However, note that these methods are _not_ provable, meaning you cannot use them in your final zkApp as a part of the ZKP. Use them only to debug your code.
+When you perform an operation inside a finite field, the result always stays in the range. This is achieved through [modular arithmetic](https://en.wikipedia.org/wiki/Modular_arithmetic). This behaviour may be unfamiliar to you if you have not worked with ZKPs before. Let us explain more with some examples.
+
+In the addition or multiplication operations where the result is bigger than the order of the finite field, you receive the modulus of the result with the prime order of the field.
```
mitschabaude, line 160:
```diff
- you receive the modulus of the result with the prime order of the field.
+ you receive the remainder of the result modulo the prime order of the field.
```

docs/zkapps/o1js/basic-concepts.md
```diff
@@ -127,25 +126,106 @@ You can think assertions in o1js similar to `require()` function in Solidity.
...
-console.log(x.toString()); // string
-console.log(x.toBigInt()); // bigint
-console.log(x.toJSON()); // JSON string
+<!-- prettier-ignore -->
+```ts
+console.log(Field.ORDER);
 ```
 
-However, note that these methods are _not_ provable, meaning you cannot use them in your final zkApp as a part of the ZKP. Use them only to debug your code.
+When you perform an operation inside a finite field, the result always stays in the range. This is achieved through [modular arithmetic](https://en.wikipedia.org/wiki/Modular_arithmetic). This behaviour may be unfamiliar to you if you have not worked with ZKPs before. Let us explain more with some examples.
```
mitschabaude, line 158: this is great!

docs/zkapps/o1js/basic-concepts.md
```diff
@@ -127,25 +126,106 @@ You can think assertions in o1js similar to `require()` function in Solidity.
...
-## Other Types and Methods
+If you try creating a `Field` from a negative number you will receive the modular additive inverse of the number. For instance `(1 + (Field.ORDER - 1)) % Field.ORDER` is 0, so the additive inverse of 1 is `Field.ORDER - 1` in this field.
 
-In addition to the `Field` type, o1js gives you a variety structures to help you build a powerful zkApp. You can learn more about them in the next chapter, [Common Types and Functions](./common-types-and-functions.md).
+<!-- prettier-ignore -->
+```ts
+Provable.log(Field(-1)); // This will print `Field.ORDER - 1`
+```
+
+Similarly, if your substraction results in a negative number you will get the modular additive inverse.
```
mitschabaude, line 174: it's confusing to me that you describe negative numbers / subtraction results in different language than positive overflow. it's the same so should be described in the same language:

"Similarly, if your subtraction results in a negative number, you receive the positive remainder of the result modulo the prime order of the field."

docs/zkapps/o1js/basic-concepts.md
```diff
@@ -127,25 +126,106 @@ You can think assertions in o1js similar to `require()` function in Solidity.
...
+
+For division, if the result is not an integer, you receive the [modular multiplicative inverse](https://en.wikipedia.org/wiki/Modular_multiplicative_inverse) of the division: If `(x * y) % p` is `z` (where `p` is the order of the finite field), then the modular multiplicative inverse of the division `z / x` equals `y`. Note that `x`, `y`, and `z` are all integers.
```
mitschabaude, line 176:
```diff
- For division, if the result is not an integer, you receive the 
+ For division, you receive the 
```
of course, if the result is an integer, it's still the modular inverse

docs/zkapps/o1js/basic-concepts.md
```diff
@@ -127,25 +126,106 @@ You can think assertions in o1js similar to `require()` function in Solidity.
...
+Provable.log(Field(3).div(Field(5))); // This will print a very big integer
+```
+
+Because of this behaviour, it is usually not very useful to use the final result of an uneven division. Instead, you can prove the result with multiplication.
+
+<!-- prettier-ignore -->
+```ts
+// We want x, an argument of the smart contract, to be equal to 7 / 3
+x.mul(Field(3)).assertEquals(Field(7));
+```
```
mitschabaude, line 189: I don't see how this is "more useful" than doing `x.div(3)`

docs/zkapps/o1js/basic-concepts.md
```diff
@@ -127,25 +126,106 @@ You can think assertions in o1js similar to `require()` function in Solidity.
...
+Provable.log(Field(-1)); // This will print `Field.ORDER - 1`
+```
+
+Similarly, if your substraction results in a negative number you will get the modular additive inverse.
+
+For division, if the result is not an integer, you receive the [modular multiplicative inverse](https://en.wikipedia.org/wiki/Modular_multiplicative_inverse) of the division: If `(x * y) % p` is `z` (where `p` is the order of the finite field), then the modular multiplicative inverse of the division `z / x` equals `y`. Note that `x`, `y`, and `z` are all integers.
+
+<!-- prettier-ignore -->
+```ts
+Provable.log(Field(3).div(Field(5))); // This will print a very big integer
```
mitschabaude, line 180: we should always use number inputs to methods instead of `Field(number)`, looks cleaner:
```ts
Provable.log(Field(3).div(5)); // This will print a very big integer
```
this applies to the entire PR!

docs/zkapps/o1js/basic-concepts.md
```diff
@@ -127,25 +126,106 @@ You can think assertions in o1js similar to `require()` function in Solidity.
...
+x.mul(Field(3)).assertEquals(Field(7));
+```
+
+If you want to have integer division instead, you can use `UInt32` or `UInt64` types (see [Common Types and Functions](./common-types-and-functions.md)).
+
+## Provability and Provable Code in o1js
+
+We have stated many times above that all the code you write in o1js must be provable, but what is provability?
+
+A provable code can be thought as a code that o1js compiler can convert into a ZKP. Just because you write something with o1js, it does not become automatically provable. You must follow all the rules above and always use provable types in your code.
```
mitschabaude, line 197: > o1js compiler

let's not use the term "compiler", it's a confusion already that people think we compile TS to circuits

docs/zkapps/o1js/basic-concepts.md
```diff
@@ -127,25 +126,106 @@ You can think assertions in o1js similar to `require()` function in Solidity.
...
+
+However, provable types are created from regular TS variables.
+
+<!-- prettier-ignore -->
+```ts
+const x = 42; // This is TS number
+let x_provable = Field(x); // This code is perfectly valid
+```
+
+In reality, the `x_provable` variable in the example above is a _constant_ `Field` variable. By constant, we do not mean that it is a TS constant (which you define by using `const`), but we say that the value of `x_provable` is known at the compilation time. When you receive an argument to your smart contract method, then the argument is not a constant, thus you should only use provable methods on it.
```
mitschabaude, line 207: nice!

docs/zkapps/o1js/basic-concepts.md
```diff
@@ -127,25 +126,106 @@ You can think assertions in o1js similar to `require()` function in Solidity.
...
+```
+
+Always remember that o1js is used for writing ZKPs, and what you write with o1js must have a sense in the final proof. The public or private inputs of a ZKP may change depending on the user input, but once the inputs are provided, the final ZKP must make sure they are not changed. By using provable o1js variables, you make sure that your variables are always consistent with your proof logic.
+
+<!-- prettier-ignore -->
+```ts
+@method foo(x: Field, y: Field) {
+  x.add(y).assertEquals(5); // we know neither the value of x, nor of the y, but we are sure that (x + y) % Field.ORDER equals 5
+}
+```
```
mitschabaude, line 231: all this is really good and important to add, thanks @yunus433

docs/zkapps/o1js/common-types-and-functions.md
mitschabaude: I would do this:
```ts
loopEnded = power.equals(i).or(loopEnded);
```
but yours is also ok

docs/zkapps/o1js/basic-concepts.md
mitschabaude:
```suggestion
A `Field` represents a positive integer in a prime order [finite field](https://en.wikipedia.org/wiki/Finite_field), which is the set of integers $0,...,p-1$, where $p$ is the field order.

Concretely, `Field` refers to the [Pallas base field](TODO), which is the native circuit field of the proof system underlying o1js.
```

@yunus433 this is how I would describe the field

dkijania:
@yunus433 I'm cleaning old PR and this one caught my eye. Looks like it is pretty old with unadressed feedback. Are you going to fix and rebase pr or shall we close it ?

yunus433:
> @yunus433 I'm cleaning old PR and this one caught my eye. Looks like it is pretty old with unadressed feedback. Are you going to fix and rebase pr or shall we close it ?

If docs are still active and this is relevant, I can fix and rebase, it should not take too much time. Let me know

PR #852, opened 2024-02-16
fix: change o1js dependency 16 in package.json

barriebyron:


examples/zkapps/01-hello-world/package.json
```diff
@@ -43,6 +43,6 @@
     "typescript": "^4.7.2"
   },
   "peerDependencies": {
-    "o1js": "0.15.*"
+    "o1js": "0.16.*"
```
barriebyron, line 46: question for SDK reviewer: do we want to make all tutorial package.json have the `0.16.*` dependency?

Trivo25: I think we should try to keep them as up to date as possible. I wouldn
t go back and fix all of them just yet, but if we decide to improve the docs <> o1js workflow in the near future, this will be much easier!

Trivo25 (approved):

PR #859, opened 2024-02-22
Fix lightnet documentation page

satyambnsal:
Fix example Github link on lightnet documentation


barriebyron (approved):
@satyambnsal thanks for the fix! we love our contributors

barriebyron:
this fix showed up in another PR that was merged first, but we still love your fix! thank you

PR #861, opened 2024-02-22
fix: update to o1js in experimental doc

barriebyron:
This doc is not in the sidebar, so earlier updates missed this instance


jackryanservia (approved):
Good catch! :)

But I don't think this doc is referenced anywhere! It's probably fine to delete it! But, there's also nothing wrong with leaving it around so that it's there in case we find a place for it later. I'll let you decide!

PR #871, opened 2024-02-29
feat: starter CHANGELOG for docs

barriebyron:
This PR starts (belatedly) a docs repo CHANGELOG 

Entries with links to the docs that have updates our users will need to be successful with Berkeley

Links will work after docs are live


barriebyron:
@halsaphi @vfrsilva and MF friends, let's particularly list the technical docs that have changes that require action from our users

PR #918, opened 2024-04-30
Update permission docs to include verification key upgrade changes

MartinMinkov:
Addresses https://app.zenhub.com/workspaces/sdk-6130fedb3b0fc600123d8796/issues/gh/o1-labs/docs2/899


docs/zkapps/o1js/permissions.mdx
```diff
@@ -331,6 +331,31 @@ await tx.prove();
...
+
+To ensure that zkApps can upgrade their verification key after a protocol upgrade, there are special rules in place for the `Impossible` and `Proof` permissions on a verification key. Recall that the `Impossible` permission prevents the verification key from being changed, while the `Proof` permission requires a valid proof to change the verification key.
+
+Internally, verification key permissions are represented as a tuple of two fields: one being the control mechanism used to change the verification key (e.g. `Impossible`, `Signature`, `Proof`) and a transaction version field (which is denoted by an integer). The transaction version field specifies the transaction version that the zkApp's verification key is associated with, allowing the protocol to manage and enforce permissions based on the current and past versions. The transaction version field is internal and is automatically set when using o1js, so developers normally do not have to think about it. However, when a protocol upgrade occurs, the transaction version field is updated to reflect the new protocol version.
```
nicc, line 340:
```suggestion
Internally, verification key permissions are represented as a tuple of two fields: the mechanism that controls verification key changes (e.g. `Impossible`, `Signature`, `Proof`), and a transaction version field (denoted by an integer). The transaction version field specifies the version of a transaction that was supported by the protocol when the zkApp's verification key was deployed. This allows the protocol to appropriately enforce permissions when a zkApps was deployed on an earlier version that may no longer be supported. The transaction version field is internal and is automatically set when using o1js, so developers normally do not have to think about it.
```

docs/zkapps/o1js/permissions.mdx
```diff
@@ -331,6 +331,31 @@ await tx.prove();
...
+
+When a zkApp is deployed, the transaction version field is set to the current version that the protocol expects. When Mina Protocol is upgraded, the protocol will expect a version that is higher than the previous version. This is where the special rules for the `Impossible` and `Proof` permissions come into play.
```
nicc, line 342: This paragraph is now redundant and can be removed (if prior suggestions are accepted).

docs/zkapps/o1js/permissions.mdx
```diff
@@ -331,6 +331,31 @@ await tx.prove();
...
+
+To allow zkApps to upgrade their verification key after a protocol upgrade, the `Impossible` and `Proof` verification key permissions are automatically treated as `Signature` permissions when the current transaction version is lower than the one Mina Protocol expects. This allows zkApps to upgrade their verification key after a protocol upgrade, even if the permission was originally set to `Impossible` or `Proof`. Once the verification key is upgraded, the transaction version field will automatically be set and the permission will revert back to its original setting.
```
nicc, line 344:
```suggestion
To allow zkApps to upgrade their verification key after a protocol upgrade, the `Impossible` and `Proof` verification key permissions are automatically treated as `Signature` permissions when the zkApp's transaction version is lower than the current one on Mina Protocol. This allows zkApps to upgrade their verification key after a protocol upgrade, even if the permission was originally set to `Impossible` or `Proof`. Once the verification key is upgraded, the transaction version field will automatically be set to the latest value and permission behaviour will revert to its original setting.
```

docs/zkapps/o1js/permissions.mdx
```diff
@@ -331,6 +331,31 @@ await tx.prove();
...
+For example, consider a zkApp that was deployed with the following permissions:
+
+```ts
+this.account.permissions.set({
+  ...Permissions.default(),
+  setVerificationKey: Permissions.impossible(),
+});
+```
+
+After a protocol upgrade, the zkApp will still be able to upgrade its verification key by providing a valid signature. The zkApp can then upgrade the verification key to be compatible with the new protocol. Once the verification key is upgraded, the permission will revert back to `Impossible`. It's important to note that the transaction version field is automatically set by o1js and developers do not need to worry about it. When it comes to upgrading your zkApp after a protocol upgrade, you can simply provide a valid signature to upgrade the verification key.
```
nicc, line 355:
```suggestion
After a protocol upgrade, the zkApp will still be able to upgrade its verification key by providing a valid signature. This allows the zkApp to upgrade its verification key to be compatible with the new protocol. Once the verification key is upgraded, the permission will revert back to `Impossible`. It's important to note that the transaction version field is automatically set by o1js and developers do not need to worry about it. When it comes to upgrading your zkApp after a protocol upgrade, you can simply provide a valid signature to upgrade the verification key.
```

nicc (changes requested):
Suggested some editorial changes. Nothing major, just flow and readability stuff.

docs/zkapps/o1js/permissions.mdx
MartinMinkov: Sounds great! Other suggestions have been accepted and this paragraph has been removed!

nicc (approved):

PR #933, opened 2024-05-06
docs(anonymous-message-board): update tutorial

MartinMinkov:
Updates the tutorial to use the latest o1js version. Also updates the content in the tutorial to simplify the project creation process by specifying flags for the cli to use instead of instructing for manual selection

### DIFF
--- a/docs/zkapps/tutorials/anonymous-message-board.mdx
+++ b/docs/zkapps/tutorials/anonymous-message-board.mdx
@@ -40,27 +40,16 @@ $ npm install -g zkapp-cli
 
 This tutorial has been tested with:
 
-- [zkApp CLI](https://www.npmjs.com/package/zkapp-cli) version `0.17.2`
-- [o1js](https://www.npmjs.com/package/o1js) version `0.16.2`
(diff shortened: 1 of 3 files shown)


docs/zkapps/tutorials/anonymous-message-board.mdx
```diff
@@ -82,8 +71,7 @@ Example smart contract files come with the new project, you can delete them if y
 1. Optional: Delete the example files not used in this tutorial:
 
   ```sh
-  $ rm src/Add.ts
-  $ rm src/Add.test.ts
+  $ rm -rf src/*
```
ymekuria, line 74: Nice. We should probably use this pattern of removing everything in the src folder in all the tutorials unless the index file is used.

ymekuria (approved):
Thanks for doing this @MartinMinkov.

docs/zkapps/tutorials/anonymous-message-board.mdx
ymekuria: I meant to communicate that I think we should use the pattern you have here `rm -rf src/*` in all the tutorials that don't use the index file.

PR #936, opened 2024-05-09
Add more detail to new verification key permission

Trivo25:
mergen when https://github.com/o1-labs/o1js/pull/1639 is released


docs/zkapps/o1js/permissions.mdx
```diff
@@ -331,27 +416,6 @@ await tx.prove();
 await tx.sign([feePayerKey, zkappKey]).send();
 ```
 
-## Upgrading after an update to the Mina Protocol
```
Trivo25, line 334: moved this above and add some changes to give more context before we dive into examples

mitschabaude (approved):
I appreciate the added detail!

PR #957, opened 2024-05-15
Offchain storage docs v1

Trivo25:


mitschabaude (approved):
Very nice, some comments:
* The examples should mostly use `update()` instead of `set()`
* `set()` is renamed to `overwrite()`
* show how to import from `Experimental`

I think this is good enough without the e2e tutorial for now

PR #1006, opened 2024-06-20
Fix broken links for zkApps docs

AyushBherwani1998:
## Description

While going through the docs, I saw a few broken links for zkApps docs. Maybe we can use something like [linkspector](https://github.com/UmbrellaDocs/linkspector) to help identify if any PR is breaking the links and also previously broken links. 

We do use it for our docs. Let me know, happy to help and set up.


ymekuria (approved):
Thanks for fixing these @AyushBherwani1998!

PR #1010, opened 2024-06-25
Update docker compose example with Mainnet image and remove libp2p sc…

kaozenn:


shimkiv:
Hey @kaozenn don't we need to also explain (perhaps in code comments) how to operate within the devnet for testing purposes?

kaozenn:
It is a good point. I will update the PR

shimkiv (approved):

PR #1045, opened 2024-09-19
Improve web worker interface

ymekuria:
## Summary

This PR improves the implementation of integrating web workers in a UI with `o1js`. The [comlink](https://www.npmjs.com/package/comlink) package was introduced to simplify the communication with web workers by abstracting message passing, and making them appear like regular asynchronous functions without the complexity of managing postMessage.

## Impact

Currently web workers are necessary to build performant and usable UIs with `o1js`. Integrating and managing web workers with `o1js` is complex, and hard to manage according to many developers on discord. These developers have requested that we simplify the process. Integrating comlink simplifies the DX and allows more developers to build performant UIs with less complexity.  

## Changes

- Introduced the comlink package
- `zkappWorker.ts` and `zkappWorkerClient.ts` were simplified
- Error handling was added in the example to aid debugging
- The UI example was refactored so devs can more easily reason about it
- The project was migrated from the `NextJS` `page router` to the `app router`
- The dependancies were updated

## Future Plans

The long term goal is to eliminate the need for web workers with updates to `o1js`. In the short term we are aiming to improve the DX of building UIs with changes like this one. This example implementation will be incrementally improved until we reach the longer term goal with improvements to `o1js`.


examples/zkapps/04-zkapp-browser-ui/ui/app/zkappWorker.ts
```diff
@@ -0,0 +1,54 @@
+import { Mina, PublicKey, fetchAccount } from 'o1js';
+import * as Comlink from "comlink";
+import type { Add } from '../../contracts/src/Add';
+
+type Transaction = Awaited<ReturnType<typeof Mina.transaction>>;
+
+const state = {
+  Add: null as null | typeof Add,
```
Trivo25, line 8: nit: maybe calling this `AddContract` and renaming `zkapp` to `contractInstance` makes this a bit clearer

Trivo25 (commented):
Cool, this is a great UX improvement! I think we should amend the actual tutorial too to include a better explanation of what Comlink does and how we use it

examples/zkapps/04-zkapp-browser-ui/ui/app/zkappWorker.ts
ymekuria: That's a good idea.

ymekuria:
> Cool, this is a great UX improvement! I think we should amend the actual tutorial too to include a better explanation of what Comlink does and how we use it

I will update the actual tutorial in a follow up PR.

querolita (commented):
Looks good to me, but I am missing context about what it was like before. I can approve, but probably someone else should give you their opinion.

45930 (approved):
LGTM!

PR #1050, opened 2024-10-03
Add Github Actions Testing to Examples

boray:
This PR updates the test-tutorials workflow to run examples when a commit is pushed to the main branch.

- Bump node version to 20
- Added automated testing for the following examples:
- `01-hello-world`
- `02-private-inputs-and-hash-functions`
- `05-common-types-and-functions`
- `10-account-updates`

### DIFF
--- a/.github/workflows/test-tutorials.yml
+++ b/.github/workflows/test-tutorials.yml
@@ -3,6 +3,7 @@ on:
   push:
     branches:
       - main
+      - feat/actions
   workflow_dispatch: {}
 
 jobs:
--- a/.github/workflows/test-tutorials.yml
+++ b/.github/workflows/test-tutorials.yml
@@ -19,7 +19,7 @@ jobs:
 
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v2
+        uses: actions/checkout@v4
       
       - name: Setup Node.js
         uses: actions/setup-node@v3
--- a/.github/workflows/test-tutorials.yml
+++ b/.github/workflows/test-tutorials.yml
@@ -2,6 +2,8 @@ name: Test-Tutorials
 
 on:
   push:
+    branches:
+      - main
 
 jobs:
   run-tests:
--- a/.github/workflows/test-tutorials.yml
+++ b/.github/workflows/test-tutorials.yml
@@ -34,7 +34,7 @@ jobs:
           git config --global user.email "test@example.com"
           git config --global user.name "Test"
 
-      - name: Install and run build for ${{ matrix.example-path }}
+      - name: Install dependencies, build and run ${{ matrix.example-path }}/build/src/main.js
         run: |
           cd ${{ matrix.example-path }}
           npm ci
--- a/.github/workflows/test-tutorials.yml
+++ b/.github/workflows/test-tutorials.yml
@@ -22,7 +22,7 @@ jobs:
         uses: actions/checkout@v4
       
       - name: Setup Node.js
-        uses: actions/setup-node@v3
+        uses: actions/setup-node@v4
         with:
           node-version: '20'
       
--- a/.github/workflows/test-tutorials.yml
+++ b/.github/workflows/test-tutorials.yml
@@ -6,7 +6,7 @@ on:
       - main
 
 jobs:
-  run-tests:
+  test-tutorials:
     runs-on: ubuntu-latest
 
     strategy:
--- a/.github/workflows/test-tutorials.yml
+++ b/.github/workflows/test-tutorials.yml
@@ -26,9 +26,6 @@ jobs:
         with:
           node-version: '20'
       
-      - name: Install dependencies at root level
-        run: npm ci
-      
       - name: Set Git config (global)
         run: |
           git config --global user.email "test@example.com"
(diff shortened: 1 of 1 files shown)


.github/workflows/test-tutorials.yml
```diff
@@ -2,18 +2,41 @@ name: Test-Tutorials
...
+      - name: Install dependencies at root level
+        run: npm ci
+      
+      - name: Set Git config (global)
+        run: |
           git config --global user.email "test@example.com"
           git config --global user.name "Test"
-          npx ts-node scripts/tutorial-runner.ts docs/zkapps/tutorials/01-hello-world.mdx
+
+      - name: Install and run build for ${{ matrix.example-path }}
```
shimkiv, line 37:
```suggestion
      - name: Install dependencies, build and run ${{ matrix.example-path }}/build/src/main.js
```

.github/workflows/test-tutorials.yml
```diff
@@ -2,18 +2,41 @@ name: Test-Tutorials
...
         with:
-          # In case of version change, update README.md accordingly
-          node-version: 18
-      - run: |
-          npm ci
+          node-version: '20'
+      
+      - name: Install dependencies at root level
+        run: npm ci
+
```
shimkiv, line 31:
```suggestion
```

.github/workflows/test-tutorials.yml
```diff
@@ -2,18 +2,41 @@ name: Test-Tutorials
...
+          - examples/zkapps/01-hello-world
+          - examples/zkapps/02-private-inputs-and-hash-functions
+          - examples/zkapps/05-common-types-and-functions
+          - examples/zkapps/10-account-updates
+
     steps:
-      - uses: actions/checkout@v2
-      - uses: actions/setup-node@v3
+      - name: Checkout repository
+        uses: actions/checkout@v2
```
shimkiv, line 22:
```suggestion
        uses: actions/checkout@v4
```

.github/workflows/test-tutorials.yml
```diff
@@ -2,18 +2,41 @@ name: Test-Tutorials
...
+      
+      - name: Setup Node.js
+        uses: actions/setup-node@v3
```
shimkiv, line 25:
```suggestion
        uses: actions/setup-node@v4
```

.github/workflows/test-tutorials.yml
```diff
@@ -2,18 +2,41 @@ name: Test-Tutorials
 
 on:
   push:
+    branches:
+      - main
 
 jobs:
-  hello-world:
+  run-tests:
```
shimkiv, line 9:
```suggestion
  test-examples:
```

.github/workflows/test-tutorials.yml
```diff
@@ -2,18 +2,38 @@ name: Test-Tutorials
...
+        uses: actions/checkout@v4
+      
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
         with:
-          # In case of version change, update README.md accordingly
-          node-version: 18
-      - run: |
-          npm ci
+          node-version: '20'
```
45930, line 27: With the matrix strategy, we can run this on node 18, 20 and 22.  I think we might as well, so we know what works and doesn't work.

.github/workflows/test-tutorials.yml
```diff
@@ -2,18 +2,38 @@ name: Test-Tutorials
 
 on:
   push:
+    branches:
+      - main
```
shimkiv, line 6:
```suggestion
on:
  push:
    branches:
      - main
  workflow_dispatch: {}
```

45930 (approved):
This is a good step!  Thanks for implementing it.  I left one suggestion about using multiple node versions for broader coverage.

.github/workflows/test-tutorials.yml
boray: It would be good to check, but do you think it’s worth the 3x time cost?

shimkiv: Jobs that use matrix will be executed in parallel, so no extra cost.

.github/workflows/test-tutorials.yml
boray: There was already a `name: Test-Tutorials` in the first line. I'll remove the one added with this suggestion. Does it work for you?

boray: Also, run workflow doesn't show up in Actions. Are you able to run it?

shimkiv: These changes have to be merged into the `main` first to let the workflow be manually triggered.

shimkiv (approved):

.github/workflows/test-tutorials.yml
shimkiv: If you want to make this workflow to execute on every push and PR then it should be something like:

```yaml
on:
  push: {}
  pull_request: {}
  workflow_dispatch: {}
```

boray: I don’t think it should be executed on every push and PR, as changes in doc2 won’t affect how examples run. Running it on every push to main is sufficient, IMO.

PR #1051, opened 2024-10-08
Bump dependencies of examples

boray:
This PR upgrades dependencies of the examples to the latest versions and removes mina-transaction-visualizer package which is out-of-date.

Dependencies upgraded:
- 01-hello-world
- 02-private-inputs-and-hash-functions
- 04-zkapp-browser-ui
- 05-common-types-and-functions
- 07-oracles
- 09-recursion
- 10-account-updates
- anonymous-message-board
- interacting-with-zkApps-server-side


45930:
@boray what was your process for doing these upgrades, did you just do them by hand or use a script?  Might be nice to do this automatically once in a while if we can also rely on CI to break on breaking changes.

45930:
I will run through all of the examples locally to confirm the new versions work, then approve

45930 (approved):

PR #1057, opened 2024-10-15
Adding upgradability to features overview

45930:
Adding a short article about upgradability to the `zkapps/feature-overview` section of the docs.

This is inspired by DFST's example code shared here: https://discord.com/channels/484437221055922177/1207271738388516884/1291820682828710021

https://github.com/zkcloudworker/zkcloudworker-tests/blob/main/tests/vk.test.ts

I realized we don't have an example showing how to actually upgrade a smart contract in the docs.  Only references to the permission to do so.

In addition to the article, I added a new directory in the examples called `feature-overview`, which has an `npm test` script that we can run in CI.  Other snippets of feature overview code could be added there to be included in CI as well.

### DIFF
--- a/docs/zkapps/writing-a-zkapp/feature-overview/upgradability.mdx
+++ b/docs/zkapps/writing-a-zkapp/feature-overview/upgradability.mdx
@@ -100,3 +100,103 @@ await tx.send();
 ```
 
 And just like that, when the transaction is applied to the blockchain, the verification key at the ZkApp address will be updated from `"27729068461170601362912907281403262888852363473424470267835507636847418791713"` to `"18150279532259194644722165513074833862035641840431153413486908511595437348455"`!
+
+## Upgrades with More Surface Area
+
+In the previous example, we changed one small variable in the smart contract but kept the state and methods the same.  However, there is no limit to what you can change via upgrade.  Let's go through one more example where more things have changed.
+
+```ts
+import { Field, SmartContract, state, State, method } from 'o1js';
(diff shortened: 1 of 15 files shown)


boray:
The example code isn’t linked in the document. Why not include a reference?

45930:
@boray it's something I considered, but I think it distracts from the main point.  The code is slightly different from the example, and it's not all necessary to understand the main point of the article.  We could add it someday if we think it's needed.

I included it as a test anyways because that will still give us an alert if the example becomes outdated.

docs/zkapps/writing-a-zkapp/feature-overview/upgradability.mdx
```diff
@@ -0,0 +1,202 @@
...
+const verificationKey = (await AddV2.compile()).verificationKey;
+const contractAddress = zkAppKey.toPublicKey();
+
+const upgradeTx = await Mina.transaction({ sender, fee },
+  async () => {
+    const update = AccountUpdate.createSigned(contractAddress);
+    update.update.verificationKey = {
+      isSome: Bool(true),
+      value: verificationKey,
+    };
```
mitschabaude, line 95: there's a nicer way to interact with the `update` field of an account update:
```
update.account.verificationKey.set(verificationKey);
```

in general documented here: https://docs.minaprotocol.com/zkapps/writing-a-zkapp/feature-overview/on-chain-values#setting-account-fields

PR #1058, opened 2024-10-15
Fix tutorials test-features CI

45930:
Fixes a CI failure on `main` from my last commit


45930:
@boray - Is there a reason not to run the CI on PRs as well as on push to main?

boray:
> well

yes, to save Github Actions cost. But, you can manually run from Actions tab.

boray:
> @boray - Is there a reason not to run the CI on PRs as well as on push to main?

Is there a reason not to run the CI on PRs as well as on push to main?

boray:
In the context of this PR, you cannot run this version of the workflow in Github Actions. But, you can still test it locally with `act`.

.github/workflows/test-tutorials.yml
```diff
@@ -3,6 +3,7 @@ on:
   push:
     branches:
       - main
+  pull_request:
```
boray, line 6: Most of the PRs are not effecting examples so we shouldn't run this workflow on every PR.

.github/workflows/test-tutorials.yml
```diff
@@ -51,7 +52,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [18, 20, 22]
+        node-version: [20, 22]
```
boray, line 55: Why not tested on node v18?

45930: I thought node test runner was introduced in 20, but I guess it will work in 18 as well.  Let's find out.

boray (approved):
LGTM!

PR #1072, opened 2024-11-13
Update zkapp-development-frameworks.mdx

JanSlobodnik:
update the doc with protokit and explanation for devs to know which one to select
This update is without the decision tree image, i need push access to add images.

### DIFF
--- a/docs/zkapps/zkapp-development-frameworks.mdx
+++ b/docs/zkapps/zkapp-development-frameworks.mdx
@@ -15,68 +15,214 @@ keywords:
   - protokit
 ---
 
-:::info
 
-To protect end users and ensure your zkApps are secure, consider the information at [Security and zkApps](/zkapps/writing-a-zkapp/introduction-to-zkapps/secure-zkapps) while architecting your solution and consider a third-party security audit before deploying to Mina mainnet.
+#  zkApp Development Frameworks
 
-:::
+Applications on Mina are called zkApps, and they come in a wide range of forms, including DeFi, games, state bridges, and more. However, due to Mina’s client execution model, the developer experience differs somewhat from that of other blockchain networks, such as Ethereum.
 
-# zkApp Development Frameworks
+Developers build zkApps with **o1js**. Developers can either use o1js directly or use o1js with **Protokit**, a framework built on top of o1js. Both options offer the full benefits of Mina’s verifiable, privacy-preserving zero knowledge proofs, but offer different tradeoffs between developer experience, performance, and decentralization.
 
-Developers building zkApps in the Mina ecosystem can leverage two different frameworks, each tailored to optimize different types of solutions.
-Explore the options below to find the perfect fit for your project.
+<table>
+  <tr>
+   <td>Info
+   </td>
+  </tr>
+  <tr>
+   <td>Please note that **Protokit is in alpha**, and settlement support with reorgs is still in progress.
+<p>
+   </td>
+  </tr>
+</table>
+
+
+This page provides a guide to choosing the best approach for your zkApp development - o1js directly or Protokit on top of o1js.
 
 If you are unsure about any of the information presented here and need guidance on choosing the most suitable framework for you, drop by [Discord](https://discord.gg/minaprotocol) and let us help!
 
-## [o1js](/zkapps/o1js)
 
-o1js is the framework for building **zkApps on the Mina L1** and new infrastructure such as rollups.  
-o1js is TypeScript based for ease of use, comes with a host of built-in features, is extensible to suit various use cases, and takes full advantage of the unique aspects of the Mina protocol.
+## Choosing a framework
 
-o1js is also the **zkDSL** used for:
+Developers can build zkApps either directly with **o1js**, or with **o1js** + **Protokit**.
 
-- Writing general-purpose zk circuits.
-- Constructing new primitives and data structures.
+Protokit is built on top of o1js and simplifies the development of **zkApps with concurrent users requiring a shared state**, such as DEXes or games. Protokit is recommended for most developers.
 
-There are some key considerations when choosing to build a zkApp with o1js on Mina L1:
+O1js is the best way to write **custom zero knowledge circuits** and construct low-level primitives on Mina, such as an NFT standard or a zkML library.
 
-- zkApps are subject to protocol throughput limitations.
-- At present, zkApps that require support for multiple concurrent users require specific architecture to avoid race conditions:
-  - Where more than the eight on-chain field elements are required to manage state, and access to that state is not shared between users, the experimental [Offchain Storage API](/zkapps/writing-a-zkapp/feature-overview/offchain-storage) offers a solution.
-  - Where concurrent access to _shared global state_ is required, the required architecture is available **out of the box** when using the Protokit framework to build your zkApp as an zkApp-chain (L2). There is currently no easy-to-use equivalent for shared state in o1js L1 contracts.
 
-Start here:
+<table>
+  <tr>
+   <td>Info
+   </td>
+  </tr>
+  <tr>
+   <td>For developers coming from the Ethereum ecosystem, think of zkApps like smart contracts that can be verified off-chain. 
+<p>
+This enables Mina to employ a client execution model, which gives Mina its scalability and privacy properties, but it can make it difficult to develop applications with concurrent users requiring a shared state. Protokit builds on top of o1js to solve these concurrency challenges.
+   </td>
+  </tr>
+</table>
 
-- [Developer documentation](/zkapps/o1js)
-- [o1js repository](https://github.com/o1-labs/o1js)
-- [Discord](https://discord.gg/minaprotocol)
 
-## [Protokit](https://protokit.dev/)
 
-Protokit is a powerful framework designed to build **ZK appchains and smart contracts** that are user-facing, privacy-preserving, interoperable.  
-It offers a familiar developer experience similar to Solidity dApps, making it easier and intuitive for developers to leverage ZK in their blockchain solutions.
+## Protokit[​](https://docs.minaprotocol.com/zkapps/zkapp-development-frameworks#protokit)
 
-It provides a full set of tools necessary for:
+Protokit is designed for **zkApps with concurrent users requiring a shared state**, for example DEXes and most games. In addition to this Protokit provides an integrated verifiable data storage solution that further simplifies the developer experience compared to o1js. 
 
-- zkApps that require high throughput or multiple concurrent users.
-- zkApps that require shared or global state access.
-- Developers familiar with execution environments such as EVM.
-- Developers who wish to leverage the modular architecture of Protokit.
+Protokit can operate in two modes: Based Sequencing and Hybrid Sequencing.
 
-Start here:
+In **Based Sequencing mode**, Protokit sends all transactions to the Mina L1, providing the decentralization guarantees of the L1, but limited to the L1’s throughput and latency. Based Sequencing mode is best for zkApps where censorship resistance and liveness are more important than the additional performance benefits.
 
-- [Developer documentation](https://protokit.dev/docs/what-is-protokit)
-- [Protokit repository](https://github.com/proto-kit)
-- [Starter Kit](https://github.com/proto-kit/starter-kit)
-- [Discord](https://discord.gg/bEGZTWRy)
+In **Hybrid Sequencing mode**, Protokit runs an off-chain sequencer with off-chain data availability, providing higher throughput and lower latency than the Mina L1 but at the expense of decentralization. Hybrid Sequencing mode is best for zkApps where performance is critical.
 
-## Framework comparison
+**Important:** Users of a zkApp in Hybrid Sequencing mode can always **submit transactions directly to the L1**, for example, to withdraw funds (in case the zkApp sequencer is unavailable) but it does rely on the Protokit’s data availability (DA). Solutions are in the works for DA guarantees even in Hybrid Sequencing mode so that developers can get higher throughput and latency while retaining censorship resistance and liveness guarantees.
 
-|| o1js SmartContract | Protokit |
-|--|--|--|
-|**Production readiness**|v1.0 released, internal audit complete, 3rd party audit in progress.|Beta release, internal audit in progress, 3rd party audit not started. Testnet only.|
-|**Censorship resistance**|Decentralized and censorship resistant.|Censorship resistance via hybrid sequencing model.|
-|**Support for multi-user apps**|Many multi-user use cases require sophisticated architecture and are limited by L1 throughput.|Capable of handling higher throughput and multiple concurrent users, thanks to Protokit's modular sequencer.|
-|**Execution environment**|Proving off-chain, verification on-chain, transaction ordering possible on-chain.|Hybrid execution model, both on-chain (sequencer) and off-chain thanks to recursive zk-proofs, verification on-chain (MINA L1).|
-|**DX**|New programming model, distinct from traditional web3.0 development.|Module oriented app-chain development, similiar to Substrate Pallets, Cosmos SDK Modules or EVM smart contracts.|
-|**Composability**|Fully composable. Contracts can call other contracts directly within a single transaction.|Protokit supports bi-directional L2 &harr; L1 messaging out of the box.|
+Start developing zkApps with Protokit here:
+
+
+
+* [Developer documentation](https://protokit.dev/docs/what-is-protokit)
+* [Protokit repository](https://github.com/proto-kit)
+* [Starter Kit](https://github.com/proto-kit/starter-kit)
+
+
+## o1js[​](https://docs.minaprotocol.com/zkapps/zkapp-development-frameworks#o1js)
+
+[o1js](https://docs.minaprotocol.com/zkapps/o1js) is the best way to author **highly-optimized ZK circuits and primitives**, but it does not provide an easy-to-use solution for managing a shared global state.
+
+o1js can be used directly on the Mina L1, and is thus limited by the L1’s throughput and latency, but maintains strong decentralization guarantees.
+
+
+<table>
+  <tr>
+   <td>Info
+   </td>
+  </tr>
+  <tr>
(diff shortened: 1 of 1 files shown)


docs/zkapps/zkapp-development-frameworks.mdx
```diff
@@ -15,68 +15,214 @@ keywords:
   - protokit
 ---
 
-:::info
 
-To protect end users and ensure your zkApps are secure, consider the information at [Security and zkApps](/zkapps/writing-a-zkapp/introduction-to-zkapps/secure-zkapps) while architecting your solution and consider a third-party security audit before deploying to Mina mainnet.
+#  zkApp Development Frameworks
 
-:::
+Applications on Mina are called zkApps, and they come in a wide range of forms, including DeFi, games, state bridges, and more. However, due to Mina’s client execution model, the developer experience differs somewhat from that of other blockchain networks, such as Ethereum.
```
boray, line 21: nit: "client side execution model"

docs/zkapps/zkapp-development-frameworks.mdx
```diff
@@ -15,68 +15,214 @@ keywords:
...
-Developers building zkApps in the Mina ecosystem can leverage two different frameworks, each tailored to optimize different types of solutions.
-Explore the options below to find the perfect fit for your project.
+<table>
+  <tr>
+   <td>Info
+   </td>
+  </tr>
+  <tr>
+   <td>Please note that **Protokit is in alpha**, and settlement support with reorgs is still in progress.
+<p>
```
boray, line 32: no closing tag for <p>

docs/zkapps/zkapp-development-frameworks.mdx
```diff
@@ -15,68 +15,214 @@ keywords:
...
+
+<table>
+  <tr>
+   <td>Info
+   </td>
+  </tr>
+  <tr>
+   <td>For developers coming from the Ethereum ecosystem, think of zkApps like smart contracts that can be verified off-chain. 
+<p>
+This enables Mina to employ a client execution model, which gives Mina its scalability and privacy properties, but it can make it difficult to develop applications with concurrent users requiring a shared state. Protokit builds on top of o1js to solve these concurrency challenges.
```
boray, line 60: nit: client side

docs/zkapps/zkapp-development-frameworks.mdx
```diff
@@ -15,68 +15,214 @@ keywords:
...
-  - Where more than the eight on-chain field elements are required to manage state, and access to that state is not shared between users, the experimental [Offchain Storage API](/zkapps/writing-a-zkapp/feature-overview/offchain-storage) offers a solution.
-  - Where concurrent access to _shared global state_ is required, the required architecture is available **out of the box** when using the Protokit framework to build your zkApp as an zkApp-chain (L2). There is currently no easy-to-use equivalent for shared state in o1js L1 contracts.
 
-Start here:
+## o1js[​](https://docs.minaprotocol.com/zkapps/zkapp-development-frameworks#o1js)
 
-- [Developer documentation](/zkapps/o1js)
-- [o1js repository](https://github.com/o1-labs/o1js)
-- [Discord](https://discord.gg/minaprotocol)
+[o1js](https://docs.minaprotocol.com/zkapps/o1js) is the best way to author **highly-optimized ZK circuits and primitives**, but it does not provide an easy-to-use solution for managing a shared global state.
```
boray, line 90: `OffchainState` provides an out-of-the box solution to this. It's worth mentioning here.

docs/zkapps/zkapp-development-frameworks.mdx
```diff
@@ -15,68 +15,214 @@ keywords:
...
+This section provides a simple decision tree x`and comparison table to help zkApp developers choose which framework to use.
+
+![Decision Tree for zkApp devs](/img/decision-tree.png)
+
+Are you developing a zkApp with concurrent users requiring a shared state?
+1. Yes - Protokit
+    1. Are you willing to trade off L1 censorship resistance and liveness guarantees for high throughput and low latency?*
+        1. Yes - Hybrid mode
+        2. No - Based mode
+2. No - o1js
```
boray, line 129: Existence of `OffchainState` makes this tree a bit complicated than this. It's true that `OffchainState` is not scalable as base mode Protokit since it doesn't keep a db thus recreates the offchain state Merkle tree on-the-fly which creates an overhead. But, It's ready to use now (not production-ready) and worth mentioning in the tree.

docs/zkapps/zkapp-development-frameworks.mdx
```diff
@@ -15,68 +15,214 @@ keywords:
...
-:::info
 
-To protect end users and ensure your zkApps are secure, consider the information at [Security and zkApps](/zkapps/writing-a-zkapp/introduction-to-zkapps/secure-zkapps) while architecting your solution and consider a third-party security audit before deploying to Mina mainnet.
+#  zkApp Development Frameworks
 
-:::
+Applications on Mina are called zkApps, and they come in a wide range of forms, including DeFi, games, state bridges, and more. However, due to Mina’s client execution model, the developer experience differs somewhat from that of other blockchain networks, such as Ethereum.
 
-# zkApp Development Frameworks
+Developers build zkApps with **o1js**. Developers can either use o1js directly or use o1js with **Protokit**, a framework built on top of o1js. Both options offer the full benefits of Mina’s verifiable, privacy-preserving zero knowledge proofs, but offer different tradeoffs between developer experience, performance, and decentralization.
```
Trivo25, line 23: > Both options offer the full benefits of Mina’s verifiable, privacy-preserving zero knowledge proofs

Do they actually? Natively, afaik Protokit executes business logic on a server/"on-chain" - away from the client - this would render __privacy-preserving__ not true in the default case

docs/zkapps/zkapp-development-frameworks.mdx
45930: I don't think we should highlight offchain state as a serious way to handle app state until it's out of the experimental namespace.  I don't think we should make casual references to things in the docs until we're ready to stand behind them 100%.

docs/zkapps/zkapp-development-frameworks.mdx
45930: An example from the [protokit docs](https://protokit.dev/docs/runtime/methods#proofs-as-arguments):

```ts
@runtimeModule()
export class Balances extends RuntimeModule<Record<string, never>> {
  @state() public balances = StateMap.from(PublicKey, UInt64);
 
  @runtimeMethod()
  public mint(canMintProof: CanMintProof, amount: UInt64) {
    canMintProof.verify();
    this.balances.set(address, amount);
  }
}
```

Seems like using a proof input would be a valid way of preserving privacy on protokit.

45930 (commented):
@JanSlobodnik 

Will the protokit documentation be moving to the mina protocol docs?  The part about based vs. hybrid sequencing seems like it could be its own article in the protokit docs.

Overall, what message is this change intended to deliver?  I think as-written it's in quite a confusing state where it pretty explicitly recommends protokit, but still makes references to o1js to be polite.  If the goal is simply to say "protokit is the app development platform for Mina", this halfway approach seems too weak.  If the goal is to say "o1js and protokit are two app development platforms for Mina", then I think we should remove recommendations for one or the other, and stick to the facts.  I also think a factual approach will scale better when other frameworks are added.

Also, should we expect more similar PRs for docs pages as part of an effort to clean up the marketing, or is this a one-off?

docs/zkapps/zkapp-development-frameworks.mdx
Trivo25: That's not the default tho whereas with o1js it is, because everything happens locally on the clients machine

Trivo25: There's a difference between opt-in and opt-out

andrewferrone:
@45930 yes the idea is to move the Protokit documentation to the Mina Protocol docs in the future.

This update is intended to address one of the most common questions we've heard from devs when confronted with o1js and Protokit - which should I use?

The intent was to land a few core messages:

1. It's not o1js *or* Protokit, but rather o1js directly or Protokit on top of o1js.
2. Protokit provides an "out-of-the-box" solution for developing apps across shared state. What types of apps are those? DEXes, lending protocols, games, most apps.
3. Protokit has tradeoffs though (decentralization vs. throughput and latency) and can be configured in two ways - Based and Hybrid modes - depending on the dev's requirements for decentralizations vs. throughput and latency.

The goal is to help devs choose whether to develop on o1js directly or Protokit on top of o1js. If you see specific areas that are confusing let's definitely clear them up.

45930:
@andrewferrone Thanks for the reply!

> one of the most common questions we've heard from devs when confronted with o1js and Protokit - which should I use?

I see!  Yeah, it makes sense that people are confused by this, but I don't think there is a simple answer.

> It's not o1js or Protokit, but rather o1js directly or Protokit on top of o1js.

I love this message, no issues with this from my side

> Protokit provides an "out-of-the-box" solution for developing apps across shared state. What types of apps are those? DEXes, lending protocols, games, most apps.

This seems overly prescriptive to me and it also puts a narrow view on the potential use cases for Mina.  What if this message was re-written more as a comparison table that explains the properties of each option?  I suppose the difference between this format and a decision tree is that it's more ambiguous and doesn't lead to a specific recommendation, but it gives devs a lot more direction about how to start making their decision.

Maybe this is too many columns, but something like this, e.g.

|                             | o1js circuits only | o1js on Mina | Protokit on Mina (Based) | Protokit on Mina (Hybrid) |
|-----------------------------|--------------------|--------------|--------------------------|---------------------------|
| Throughput                  |                    |              |                          |                           |
| Latency                     |                    |              |                          |                           |
| Centralization              |                    |              |                          |                           |
| Infrastructure Requirements |                    |              |                          |                           |
| ...                         |                    |              |                          |                           |

docs/zkapps/zkapp-development-frameworks.mdx
```diff
@@ -15,68 +15,214 @@ keywords:
...
-Start here:
+## o1js[​](https://docs.minaprotocol.com/zkapps/zkapp-development-frameworks#o1js)
 
-- [Developer documentation](/zkapps/o1js)
-- [o1js repository](https://github.com/o1-labs/o1js)
-- [Discord](https://discord.gg/minaprotocol)
+[o1js](https://docs.minaprotocol.com/zkapps/o1js) is the best way to author **highly-optimized ZK circuits and primitives**, but it does not provide an easy-to-use solution for managing a shared global state.
 
-## [Protokit](https://protokit.dev/)
+o1js can be used directly on the Mina L1, and is thus limited by the L1’s throughput and latency, but maintains strong decentralization guarantees.
```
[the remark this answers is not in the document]
maht0rz: I agree with @Trivo25, we should be clear in distinguishing the terminology between "general o1js" and "smart contracts / account updates"

docs/zkapps/zkapp-development-frameworks.mdx
Trivo25: >  So privacy is indeed optional, but that's actually a feature you can leverage, which mitigates the concurrency issues of current smart contracts.

I am not saying it isn't a feature! But a big difference that users should be made aware of, especially since the default is on-chain execution.

andrewferrone:
@45930 did you see the comparison table at the bottom of the page?

We can remove the decision tree.

> What if this message was re-written more as a comparison table that explains the properties of each option? I suppose the difference between this format and a decision tree is that it's more ambiguous and doesn't lead to a specific recommendation, but it gives devs a lot more direction about how to start making their decision.
> 
> Maybe this is too many columns, but something like this, e.g.
> 
> o1js circuits only	o1js on Mina	Protokit on Mina (Based)	Protokit on Mina (Hybrid)
> Throughput				
> Latency				
> Centralization				
> Infrastructure Requirements				
> ...

docs/zkapps/zkapp-development-frameworks.mdx
JanSlobodnik: fixed

docs/zkapps/zkapp-development-frameworks.mdx
JanSlobodnik: fixed

docs/zkapps/zkapp-development-frameworks.mdx
```diff
@@ -15,68 +15,214 @@ keywords:
...
 
-o1js is the framework for building **zkApps on the Mina L1** and new infrastructure such as rollups.  
-o1js is TypeScript based for ease of use, comes with a host of built-in features, is extensible to suit various use cases, and takes full advantage of the unique aspects of the Mina protocol.
+## Choosing a framework
+
+Developers can build zkApps either directly with **o1js**, or with **o1js** + **Protokit**.
+
+Protokit is built on top of o1js and simplifies the development of **zkApps with concurrent users requiring a shared state**, such as DEXes or games. Protokit is recommended for most developers.
+
+O1js is the best way to write **custom zero knowledge circuits** and construct low-level primitives on Mina, such as an NFT standard or a zkML library.
```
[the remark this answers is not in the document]
JanSlobodnik: fixed

docs/zkapps/zkapp-development-frameworks.mdx
```diff
@@ -15,68 +15,214 @@ keywords:
...
+   <td>For developers coming from the Ethereum ecosystem, think of zkApps like smart contracts that can be verified off-chain. 
+<p>
+This enables Mina to employ a client execution model, which gives Mina its scalability and privacy properties, but it can make it difficult to develop applications with concurrent users requiring a shared state. Protokit builds on top of o1js to solve these concurrency challenges.
+   </td>
+  </tr>
+</table>
+
+
+
+## Protokit[​](https://docs.minaprotocol.com/zkapps/zkapp-development-frameworks#protokit)
```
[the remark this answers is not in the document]
JanSlobodnik: fixed

docs/zkapps/zkapp-development-frameworks.mdx
```diff
@@ -15,68 +15,214 @@ keywords:
...
+* [Developer documentation](https://docs.minaprotocol.com/zkapps/o1js)
+* [o1js repository](https://github.com/o1-labs/o1js)
+* [Discord](https://discord.gg/minaprotocol)
+
+
+## Choosing between Protokit and o1js
+
+**Still can’t decide between Protokit and o1js?** 
+
+This section provides a simple decision tree x`and comparison table to help zkApp developers choose which framework to use.
```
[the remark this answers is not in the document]
JanSlobodnik: fixed

docs/zkapps/zkapp-development-frameworks.mdx
andrewferrone: Can we consider this closed? Jan has address Florian's comment and we're not going to include mention of OffchainState.

docs/zkapps/zkapp-development-frameworks.mdx
andrewferrone: Considering this closed.

docs/zkapps/zkapp-development-frameworks.mdx
andrewferrone: Is everyone ok to leave this as is? The statement is still correct - both frameworks allow for privacy. The details of how to implement for privacy can be left for deeper level docs. The key point to land here is that you can build a privacy-preserving app using either framework.

docs/zkapps/zkapp-development-frameworks.mdx
```diff
@@ -15,68 +15,214 @@ keywords:
...
 
-## [o1js](/zkapps/o1js)
 
-o1js is the framework for building **zkApps on the Mina L1** and new infrastructure such as rollups.  
-o1js is TypeScript based for ease of use, comes with a host of built-in features, is extensible to suit various use cases, and takes full advantage of the unique aspects of the Mina protocol.
+## Choosing a framework
+
+Developers can build zkApps either directly with **o1js**, or with **o1js** + **Protokit**.
+
+Protokit is built on top of o1js and simplifies the development of **zkApps with concurrent users requiring a shared state**, such as DEXes or games. Protokit is recommended for most developers.
```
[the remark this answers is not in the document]
andrewferrone: We'll remove "Protokit is recommended for most developers."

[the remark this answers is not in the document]
JanSlobodnik: fixed (removed sentence)

docs/zkapps/zkapp-development-frameworks.mdx
```diff
@@ -15,68 +15,214 @@ keywords:
...
+O1js is the best way to write **custom zero knowledge circuits** and construct low-level primitives on Mina, such as an NFT standard or a zkML library.
+
+
+<table>
+  <tr>
+   <td>Info
+   </td>
+  </tr>
+  <tr>
+   <td>For developers coming from the Ethereum ecosystem, think of zkApps like smart contracts that can be verified off-chain.
```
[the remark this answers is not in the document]
JanSlobodnik: fixed

docs/zkapps/zkapp-development-frameworks.mdx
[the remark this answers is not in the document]
andrewferrone: @Trivo25 @maht0rz how would you recommend changing this sentence? We're trying to make the point here that o1js has to use L1 sequencing and therefore is constrained by L1's throughput and latency but with strong decentralization guarantees.

docs/zkapps/zkapp-development-frameworks.mdx
```diff
@@ -15,68 +15,214 @@ keywords:
...
+
+## Choosing between Protokit and o1js
+
+**Still can’t decide between Protokit and o1js?** 
+
+This section provides a simple decision tree x`and comparison table to help zkApp developers choose which framework to use.
+
+![Decision Tree for zkApp devs](/img/decision-tree.png)
+
+Are you developing a zkApp with concurrent users requiring a shared state?
```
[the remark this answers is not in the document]
andrewferrone: We'll remove the decision tree and leave the table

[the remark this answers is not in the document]
JanSlobodnik: removed decision tree, only table remains

docs/zkapps/zkapp-development-frameworks.mdx
rpanic: Its not true that OffchainState is a solution to race conditions on "shared global state" as suggested above. It only solves this for user-specific state. The thing is that you can't concurrently compute over shared global state without any form of on-chain execution. That is why we built protokit, to add that possibility to the features already existing in o1js.
If you wanted to do the same with vanilla-o1js features and then polish the DX, solve other limitations in the process and then make the system efficient, you'd pretty much end up with protokit in based sequencing mode (which is the first mode we will ship)

PR #1074, opened 2024-11-15
Update copyright on the docs footer

halsaphi:
Update to footer to have the correct copyright date.

Adding a getdate function and using it to always have correct copyright date.

### DIFF
--- a/src/theme/Footer/index.tsx
+++ b/src/theme/Footer/index.tsx
@@ -15,12 +15,6 @@ import WeChatLogo from '@site/static/svg/socials/wechat_24x24.svg';
 import YoutubeLogo from '@site/static/svg/socials/youtube_24x24.svg';
 import { FormSubmitUrl } from '@site/src/constants';
 
-function getCurrentYear(){
-  const today = new Date();
-  const year = today.getFullYear();
-  return year;
-}
-
 function Footer(): JSX.Element | null {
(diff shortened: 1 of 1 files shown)


45930 (changes requested):
I tested this locally and found that even serving the static build results in this js function running on every page load.  It's not a huge impact on performance, but it seems like unnecessary bloat.

Is there a way to get this value to be set at build time, then it will stay up to date over time but not trigger JS in the client?

halsaphi:
I'm not aware of a way to do this in Docusaurus - though I'm no expert, so if you can find a way please let me know - other option is to set it to 2024 - though this would require an update next year.

45930:
@halsaphi So one thing you can do is export a new constant from `src/constants.ts` and set it to '2024'.  Then either add a ci rule that tests the constant against the current year, or write a prebuild script (just some js that gets run automatically before build) to check the value and re-write the file if it's wrong.

I don't think updating it manually once per year is a bad idea either.

halsaphi:
@45930 - hardcoded the value -> can you merge so docs have correct copyright notice. This will need updating again in a few months.

snaitmouloud-mf:
Hi guys,
can we get this merged please ?
It's 5 days this is open just to change the year.
Thanks.

PR #1088, opened 2025-01-10
Add documentation for the Indexed Merkle Map API

Shigoto-dev19:


docs/zkapps/o1js/indexed-merkle-map.mdx
```diff
@@ -0,0 +1,263 @@
...
+// Synchronize the off-chain Indexed Merkle Map to match the on-chain state root
+indexedMerkleMap.update(Field(1), Field(5678));
+
+console.log(
+  indexedMerkleMap.root.toBigInt() === contract.mapRoot.get().toBigInt()
+);
+console.log(indexedMerkleMap.data.get().sortedLeaves);
+```
+
+## Indexed Merkle Map - API reference
```
45930, line 170: I feel strongly that this should not be included here.  This definitely belongs in the Typedoc reference, so we should work on generating that correctly instead of manually adding auto-generated type declarations here.

45930 (commented):
I think it's good to have this article alongside the typedoc, but I'd personally prefer to highlight ZkProgram rather than smart contract.

I'm also open to just promoting the API out of experimental, especially with this article written.  Let's see what @Trivo25 thinks 🤔 , so if we do that, we can also remove some of the `Experimental` references here.

docs/zkapps/o1js/indexed-merkle-map.mdx
```diff
@@ -0,0 +1,263 @@
...
+In this example, `IndexedMerkleMap31` is a Merkle map capable of holding up to 2<sup>(31−1)</sup> leaves; approximately 1 billion entries.
+
+### Utilizing IndexedMerkleMap in a smart contract
+
+Developers can integrate the `IndexedMerkleMap` into their zkApps by passing a Merkle map as a method parameter and invoking its methods as needed.
+
+This is possible because the `IndexedMerkleMap` can be used within provable code.
+
+```ts
+class MyContract extends SmartContract {
```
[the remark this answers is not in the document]
Shigoto-dev19: Regardless of documentation maintainability, I think that demonstrating a `SmartContract` example for the `Indexed Merkle Map` API provides greater benefits for developers and serves as more effective documentation. 

When someone wants to use the Indexed Merkle Map API, they will likely integrate it directly within a smart contract, as it works seamlessly in provable code. 

In contrast, a `ZkProgram` example might lead readers to mistakenly believe that the API should be used with a `ZkProgram`. This misunderstanding could imply that verifying a zkProgram proof for Indexed Merkle Map updates is straightforward, when in fact it is only practical in advanced scenarios such as recursive reducers.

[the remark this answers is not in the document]
45930: > This misunderstanding could imply that verifying a zkProgram proof for Indexed Merkle Map updates is straightforward, when in fact it is only practical in advanced scenarios such as recursive reducers.

Maybe I don't understand this myself.  What's wrong with this example:

```ts
import { Experimental, Field, SelfProof, ZkProgram } from 'o1js';

const height = 5;
class MyMerkleMap extends Experimental.IndexedMerkleMap(height) {}

const RecursiveProgram = ZkProgram({
  name: 'RecursiveProgram',
  publicOutput: MyMerkleMap,
  methods: {
    init: {
      privateInputs: [],
      method: async () => {
        const map = new MyMerkleMap();
        return { publicOutput: map };
      },
    },
    insert: {
      privateInputs: [SelfProof, Field, Field],
      method: async (
        proof: SelfProof<undefined, MyMerkleMap>,
        key: Field,
        value: Field
      ) => {
        proof.verify();
        const newMap = proof.publicOutput.clone();
        newMap.insert(key, value);
        return { publicOutput: newMap };
      },
    },
    update: {
      privateInputs: [SelfProof, Field, Field],
      method: async (
        proof: SelfProof<undefined, MyMerkleMap>,
        key: Field,
        value: Field
      ) => {
        proof.verify();
        const newMap = proof.publicOutput.clone();
        newMap.update(key, value);
        return { publicOutput: newMap };
      },
    },
  },
});

await RecursiveProgram.compile();

class RecursiveProof extends ZkProgram.Proof(RecursiveProgram) {}

const p = await RecursiveProgram.init();
const p2 = await RecursiveProgram.insert(p.proof, Field(1), Field(2));
const p3 = await RecursiveProgram.update(p2.proof, Field(1), Field(3));

console.log(p3.proof.publicOutput.get(Field(1))); // 3
```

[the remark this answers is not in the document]
Shigoto-dev19: There's nothing wrong with the example but I think whoever wants to use the `Indexed MM` would use it directly within a zkApp as it's more practical and for that, I think a `SmartContract` example would be a better guide for the reader and less confusing.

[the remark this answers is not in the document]
45930: > I think whoever wants to use the Indexed MM would use it directly within a zkApp as it's more practical

I'm not sold on this.  Do you actually believe indexed merkle map is a usable solution for zkapps today?  To me, your example is misleading because it teases the existence of a `map` API that works on Mina, but doesn't address the same old concurrent update or state storage issues that exist for every zkapp.  In production, this example would not work with multiple users accessing it at once, and I want to stop creating _new_ examples that break down at scale.

I know that this is a touchy issue in the community right now as protokit devs are looking for L1 alternatives to the `StateMap` so that they can deploy on mainnet: ([discord example](https://discord.com/channels/484437221055922177/1277461243627700325/1325894605232930858)).  IMO there is absolutely no alternative that currently works, and I don't want to beat around the bush with that fact and pretend this is a viable alternative.  Do you disagree with this?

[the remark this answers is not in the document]
Shigoto-dev19: Fair enough, I don't disagree! 

I see your point here! But that's also applicable to many primitives in the documentation. I was trying to keep it consistent with the current o1js docs as I was documenting the `Indexed Merkle Map` API.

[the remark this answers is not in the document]
45930: My goal with example code in the docs is for developers to grasp the concept and then take it to build one step further.  My problem with the pattern we have in most of our existing documentation is that it only works to a certain point, and that's what we show.  When someone tries to take the next step, they hit a wall.

What I like about `ZkProgram` is that the final product can be a Mina ZkApp, or it could live on protokit or zeko.  And before deploying it to any network, a developer can change different variables, combine it with other `ZkProgram`s, and extend it in all kinds of ways locally.

docs/zkapps/o1js/indexed-merkle-map.mdx
```diff
@@ -0,0 +1,79 @@
...
+Similar to a Merkle Tree, a Merkle Map allows referencing off-chain data by storing a single hash, also known as the root.
+
+A Merkle Map is a wrapper around a [Merkle Tree](/zkapps/o1js/merkle-tree). Both data structures are analogous, but instead of using an index to set a leaf in a tree, a Merkle Map uses a key in a map.
+
+## Design
+
+The Indexed Merkle Map is an improved version of the [MerkleMap](/zkapps/tutorials/common-types-and-functions#merkle-map), offering enhanced efficiency and usability:
+
+- **Reduced Constraints:** Uses 4-8x fewer constraints than `MerkleMap`.
+- **Support for Dummy Updates:** Handles dummy updates for keys like `0` and `-1`.
```
hattyhattington17, line 35: Could we expand on what a dummy update is?

docs/zkapps/o1js/indexed-merkle-map.mdx
```diff
@@ -0,0 +1,79 @@
...
+- **Provable Code Integration:** Unlike `MerkleTree` and `MerkleMap`, the high-level API of `IndexedMerkleMap` is usable within provable code.
+- **Comprehensive Methods:** Comes with a rich set of methods that enable a wide range of operations.
```
hattyhattington17, line 37: I'd skip this line because it adds "fluff" without really providing new info!

docs/zkapps/o1js/indexed-merkle-map.mdx
Shigoto-dev19: It's described in detail in the `Additional Resources` section and specifically in the link for `PR#1671`.
I think it's enough, or do you think we should still add more details?

hattyhattington17: I think if we mention dummy updates explicitly we should be clear about what the term means since I had to dig around in the docs a bit before I understood that it was a matter of the update not affecting the root

docs/zkapps/o1js/indexed-merkle-map.mdx
```diff
@@ -0,0 +1,78 @@
...
+Similar to a Merkle Tree, a Merkle Map allows referencing off-chain data by storing a single hash, also known as the root.
+
+A Merkle Map is a wrapper around a [Merkle Tree](/zkapps/o1js/merkle-tree). Both data structures are analogous, but instead of using an index to set a leaf in a tree, a Merkle Map uses a key in a map.
+
+## Design
+
+The Indexed Merkle Map is an improved version of the [MerkleMap](/zkapps/tutorials/common-types-and-functions#merkle-map), offering enhanced efficiency and usability:
+
+- **Reduced Constraints:** Uses 4-8x fewer constraints than `MerkleMap`.
+- **Support for Dummy Updates:** Handles updates that don't affect the root for keys like `0` and `-1`.
```
hattyhattington17, line 35: My understanding was that a dummy update was just (0,0) 🤔 how about something like:

- **Support for -1 and 0 as keys:** Useful for creating dummy updates (with key and value both set to zero) which do not alter the map root

45930: What is the _point_ of a dummy update?  I think that is the main thing to elaborate on.  Saying we can create an update that doesn't affect the root doesn't intuitively mean anything to me.  Is the useful part that we can, for instance, use it with `Provable.if` to sometimes set a value, and sometimes not?  Are there other useful cases?  Maybe we could find an intuitive case to highlight and focus on the value of a dummy update, rather than the definition?

45930 (approved):

PR #1093, opened 2025-01-22
Fix/angular integration guide

hattyhattington17:
The coi-serviceworker.min.js import wasn't working from TS so this PR updates the tutorial to import that file directly in index.html. It also makes some minor changes to project defaults, hardcodes versions of libraries, and adds images.


45930 (approved):

45930:
With updates, I am up and running! https://45930.github.io/ng-integration-guide/

PR #1111, opened 2025-06-16
Bump mina docker to 3.1.0

glyh:
The 2nd commit is reproducible with
```
./scripts/update_release_tag.sh \
  3.1.0-ae112d3\
  3.1.0-alpha1-56cdb61
```

Some sections are still outdated, e.g. git commit hash. Let me update them in subsequent PRs.

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -1,8 +1,8 @@
 {
    "release":{
       "tags":{
-         "mainnet":"3.0.3-d800da8",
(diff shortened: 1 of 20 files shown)


metadata.json
nicc, line 1: All the other changes reflect only a version bump but this file is losing information. I don't think it's a problem, just flagging in case it was an accident.

nicc (approved):

PR #1127, opened 2025-08-11
Replace all seed peers URL to the one managed by O1 instead of MF

glyh:
As MF shutdown their seed server, we manage seed node list at O1.


dannywillems (approved):

amc-ie (approved):
Thanks!

PR #1176, opened 2026-04-02
add an example docker compose for upgrading into mesa

glyh:
As title.

### DIFF
--- a/docs/network-upgrades/mesa/docker-compose-quickstart.mdx
+++ b/docs/network-upgrades/mesa/docker-compose-quickstart.mdx
@@ -15,14 +15,11 @@ keywords:
 ```yaml
 services:
   mina_node:
-    # Use the correct stop slot release auto fork docker here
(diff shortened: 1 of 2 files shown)


docs/network-upgrades/mesa/docker-compose-quickstart.mdx
```diff
@@ -0,0 +1,51 @@
...
+    restart: always
+    environment:
+      MINA_CLIENT_TRUSTLIST: "0.0.0.0/0"
+      # Your other environment variables here
+    entrypoint: []
+    # Always include `--hardfork-handling migrate-exit` so docker will use auto forking mode
+    command: >
+      bash -c '
+        mina daemon \
+             --peer-list-url https://bootnodes.minaprotocol.com/networks/mainnet.txt \
```
dkijania, line 26: comment here would be welcomed that this is for mainnet and other peer list can be delivered . This comment also can be move above the code snippet as general tip

dkijania (commented):
only one nit

dkijania (approved):

PR #1183, opened 2026-04-13
Add guide for downgrading Mina node to older versions.

glyh:
## NOTE

We haven't actually relased a breaking version until now, I think. It'd be better we test the instructions in this doc work right after the release, so to update doc accordingly.

## Summary

- Add a new doc page under Node Operators: **Downgrading to Older Versions**
- Covers how to roll back a Mina node from versions above 3.3.0 to 3.3.0 or below by converting on-disk state in place using `mina-storage-converter`, avoiding a full rebootstrap from the remote S3 ledger bucket
- Step-by-step instructions for both **Debian/Ubuntu** (stop daemon, install toolbox packages, run converter, install old package, restart) and **Docker** (stop container, run converter one-liner in the source image, start new container with target image)
- Sidebar entry added before Troubleshooting/FAQ

## Details

Node operators upgrading past 3.3.0 may later need to roll back. Without this guide, the only option is a time-consuming full rebootstrap. The `mina-storage-converter` tool can convert the local RocksDB state between versions, but the process is undocumented — this page fills that gap.

Key design choices in the guide:
- Uses `*` wildcard for the RocksDB version path so operators don't need to know which RocksDB version is bundled with each Mina release
- Uses placeholder env vars (`NODE_DIR`, `SOURCE_VERSION`, `TARGET_VERSION`) consistently across Debian and Docker sections
- Docker section provides a single `docker run ... -c "..."` one-liner that installs toolbox packages and runs the converter in one shot
- Includes a caution for Docker users whose `mina-config` is not persisted outside the container


docs/node-operators/downgrading-to-older-versions.mdx
```diff
@@ -0,0 +1,123 @@
...
+```
+
+Or however you normally stop your node process. Verify it is no longer running before continuing.
+
+### 2. Install storage toolbox packages
+
+Install the required toolbox packages that provide `mina-storage-converter` and the RocksDB scanners:
+
+```sh
+sudo apt-get install -y mina-daemon-storage-toolbox mina-daemon-recovery-storage-toolbox
```
glyh, line 36: I'm not very sure how is it packaged, exactly, this is left to reviewer to confirm.

docs/node-operators/downgrading-to-older-versions.mdx
```diff
@@ -0,0 +1,123 @@
...
+### 2. Install toolbox packages and run the converter
+
+Use the current (newer) image to install the toolbox packages and run the conversion against your mounted `mina-config` volume:
+
+```sh
+docker run -it --rm \
+    --entrypoint bash \
+    --mount "type=bind,source=$(pwd)/.mina-config,dst=/root/.mina-config" \
+    minaprotocol/mina-daemon:${SOURCE_VERSION}-bullseye-mainnet \
+    -c "apt-get update && apt-get install -y mina-daemon-storage-toolbox mina-daemon-recovery-storage-toolbox && mina-storage-converter --node-dir /root/.mina-config --current-scanner /usr/lib/mina/storage/*/${SOURCE_VERSION}/mina-rocksdb-scanner --stable-scanner /usr/lib/mina/storage/*/${TARGET_VERSION}/mina-rocksdb-scanner"
```
glyh, line 107: I'm not very sure how is it packaged, exactly, this is left to reviewer to confirm.

docs/node-operators/downgrading-to-older-versions.mdx
```diff
@@ -0,0 +1,125 @@
...
+  - mina downgrade
+  - mina storage converter
+  - node operators
+  - version rollback
+  - mina-rocksdb-scanner
+  - rebootstrap
+  - on-disk state
+---
+
+# Downgrading to Older Versions
```
dkijania, line 15: I would add some pre section for this section explaining why we ever need to rollback to previous storage version. Maybe add some example of issue was reported. Another idea would be to leave this document as it is but add another docs for known issues, where we put the issue BP were experiencing and link to this doc which should be a solution for known issues

docs/node-operators/downgrading-to-older-versions.mdx
dkijania: I think it's good, it will pull newest versions for both of them

dkijania (approved):

PR #1215, opened 2026-08-08
ci: o1js API reference workflow pins Node 16, below both projects' engines

boleklebovski:
## Problem

`.github/workflows/o1js-api-reference.yml:12` sets up Node **16**, then runs `npm ci && npm run build` against two projects that both require newer:

| Project | Declared `engines.node` |
|---|---|
| `o1-labs/o1js` (currently 2.15.0) | `>=18.14.0` |
| this repo (`docs2`) | `>=22` |

So the job installs a runtime older than either floor and then tries `npm ci` in both trees. Node 16 also reached end of life in September 2023.

The workflow is `workflow_dispatch` only, so it never blocks a PR — the effect is quieter: regenerating the o1js API reference is broken whenever a maintainer actually reaches for it.

## Fix

One line: `node-version: 16` → `node-version: 22`. That clears both floors and matches this repo's own `engines` declaration. The rest of the workflow is already current (`actions/setup-node@v6`, `actions/checkout@v6`).

The line carries the comment *"In case of version change, update README.md accordingly"* — I checked, and the README does not state a Node version anywhere, so there is nothing to keep in sync. Happy to add one if you would like the comment to stay meaningful.

## Verification

Workflow re-parsed as YAML after the edit; the `setup-node` step reads `{'node-version': 22}`. No other step, trigger or command is touched.


dkijania (approved):

dkijania:
Thanks !

Issue #1236, opened 2026-09-06
docs: request o1Labs → Mina backlinks and resolve the canonical-host question (upstream)


dkijania:
### Source
`Mina_o1js_Excel_Report.xlsx` → *Backlink Plan*: *"Current state: Mina→o1labs links on 12 pages; o1labs→Mina on only 1. The gap (o1labs→Mina) is the priority."*

### Problem
The traffic is one-way. We link out to o1Labs from 37 distinct URLs plus 20 outbound `vercel.json` redirects. o1Labs links back to Mina docs from **one** page.

A developer who starts on o1Labs has no route to the material that exists only here and is out of o1js scope by definition:

- **Network and deployment specifics** — faucet, devnet, GraphQL endpoints, `how-to-deploy-a-zkapp`, `interact-with-mina`
- **The account model, fees, mainnet behaviour**
- **zkApp CLI** — o1Labs assumes `npm install o1js`
- **Lightnet** local testing
- **Secure zkApps** guidance
- **The Mesa network upgrade** — fork schedule and the required o1js pre-release

### Requests to raise upstream in `o1-labs/o1js-documentation-site`

**1. Backlinks**

| o1Labs trigger page | Should link to |
|---|---|
| `zkapps/deployment`, `zkapps/local-development`, `zkapps/intro` | Mina: `how-to-deploy-a-zkapp`, `interact-with-mina`, network/faucet pages |
| `zkapps/*` (accounts, fees, mainnet context) | Mina: protocol concepts, `using-mina`, node basics |
| `getting-started/what-is-a-zk-constraint-system` | Mina: `getting-started-zkapps` + `install-zkapp-cli` |
| `zkapps/upgrade-mesa` | Mina: `network-upgrades/mesa` |

*(The workbook named `getting-started/intro` as the entry page; that URL now **404s**, so the constraint-system page is the live entry point.)*

**2. Canonical host mismatch — needs an answer before anything else**

Every o1Labs page declares `<link rel="canonical" href="https://documentation.o1labs.org/...">`, and the sitemap lists 1105 URLs, all on `documentation.o1labs.org`. `docs.o1labs.org` appears **zero** times in it. But `documentation.o1labs.org` **does not resolve in DNS**, while `docs.o1labs.org` serves everything with a 200 and no redirect.

[...]

Issue #1235, opened 2026-09-06
docs: two broken o1Labs links, a canonical-host mismatch, and no CI check on 57 external references


dkijania:
### Source
`Mina_o1js_Excel_Report.xlsx` → *Only in Mina*, row "FAQ": *"audit that linked anchors still exist on o1labs."* I ran that audit against the live site on 2026-09-06 and found more than the workbook anticipated.

### Finding 1 — the o1Labs site declares a canonical host that does not resolve

Every page we link to carries:

```html
<link rel="canonical" href="https://documentation.o1labs.org/o1js/zkapps/permissions">
```

and the o1Labs sitemap contains **1105 URLs, all on `documentation.o1labs.org`** — `docs.o1labs.org` appears **zero** times in it.

But `documentation.o1labs.org` **does not resolve in DNS**:

```
curl: (6) Could not resolve host: documentation.o1labs.org
```

`docs.o1labs.org` serves the content and returns 200, with no redirect. So the host we depend on works, while the host it names as canonical is dead.

**Why this matters to us:** we point **37 in-page links plus 20 `vercel.json` redirects** at `docs.o1labs.org`. Those 20 redirects are the *only* route to the 11 pages Phase 2 of #1208 deleted. If o1Labs completes the migration and retires `docs.`, all 57 break at once and 11 guides become dead ends. Search engines are meanwhile told the canonical copy lives at a nonexistent domain.

### Finding 2 — two links are broken right now

| Location | Link | Status |
|---|---|---|
| `docs/zkapps/o1js/indexed-merkle-map.mdx:73` | `.../api-reference/namespaces/Experimental/functions/IndexedMerkleMap` | **404** |
| `docs/glossary.mdx:310` | `.../zkapps/permissions#upgradeability-of-smart-contracts` | 200, **anchor gone** |

The glossary anchor is a silent failure of exactly the kind the workbook warned about: the page returns 200 but has no upgradeability section any more. Its live headings are `types-of-permissions`, `types-of-authorizations`, `authorization`, `basic-permission-setup`, `api-reference`. The upgradeability content moved to `advanced-concepts/ZkApps/upgradability`.

The remaining 35 URLs and 2 anchors (`Field#methods`, `SmartContract#requiresignature`) all resolve.

### Finding 3 — o1Labs serves versioned URLs
[...]

Issue #1208, opened 2026-05-28
Assess & stage migration of zkApp Developers section to o1Labs docs (docs.o1labs.org/o1js)


dkijania:
## Summary

We are progressively delegating o1js / zkApp content to the o1Labs docs site. This issue captures an assessment of the entire **"zkApp Developers"** section (`docs/zkapps/**`, 272 files) and a staged plan to migrate as much as possible while avoiding orphaned content and broken links.

**Verdict: do not remove the whole section at once — delegate in stages.** The clean, safe first step is retiring the auto-generated o1js **API reference** (217 of 272 files) with redirects; the hand-written guides and tutorials should stay until o1Labs reaches parity.

### Important correction on the target
`o1js.dev` is **not** the delegation target — it currently serves a stock, unmodified Docusaurus template with no real content, and `docs.o1js.org` is unreachable. The actively-maintained o1js docs live at **https://docs.o1labs.org/o1js** (o1js 2.15.0, changelog dated 2026-03-16). All redirects/banners must point there. The existing in-repo `MigrateBanner` already does.

## What's in the section

| Group | Files | Notes |
|---|---|---|
| `o1js-reference/**` (auto-generated API ref) | **217** | Machine-generated mirror of the o1js API |
| `o1js/**` (primitive guides) | 13 | basic-concepts, recursion, circuit-writing-primer, gadgets, bitwise-operations, foreign-fields, merkle-tree, indexed-merkle-map, keccak, ecdsa, sha256 |
| `writing-a-zkapp/**` | 20 | introduction-to-zkapps (11) + feature-overview (9) |
| `tutorials/**` | 13 | numbered series (hello-world … account-updates, message board, server-side) |
| `advanced/`, `front-end-integration-guides/`, top-level (`faq`, `roadmap`, `standards`, `zkapp-development-frameworks`) | ~9 | |
| **Total** | **272** | |

## Coverage: docs.o1labs.org vs. what we'd remove

**Already covered on o1labs (safe to delegate):**
- Core zkApp model: how zkApps work, smart contracts, permissions, events, actions & reducers, deployment, upgradability, local development
- o1js conceptual/primitives: constraint systems, witnesses, recursion, ZkProgram, merkle trees, hashing
- **Full generated API reference** — cleanest 1:1 overlap with our `o1js-reference/**`

**Gaps still served ONLY by Mina docs (would be orphaned by a full removal):**
- **zkApp CLI** (install, scaffolding, CLI-based deploy)
[...]

PR #1241, opened 2026-09-06
docs: cross-link three o1js pages to their o1Labs counterparts

dkijania:
Part of #1228, #1229, #1234.

Of the 43 pages left under `docs/zkapps/`, **25 carry no link to `docs.o1labs.org` at all**. This PR takes the three where an o1Labs counterpart clearly exists and is the more current reference.

| Page | Added | Why |
|---|---|---|
| `o1js/circuit-writing-primer.mdx` | `conditional-logic`, `witnesses`, `analyzing-constraint-systems`, `what-is-a-zk-constraint-system` | o1Labs splits this material across three maintained pages; our single 1531-word primer linked to none of them (#1228) |
| `o1js/basic-concepts.mdx` | the six `basic-types/*` pages + `what-is-a-zk-constraint-system` + `conditional-logic` | o1Labs is more granular on the primitives than this page is (#1229) |
| `o1js/bitwise-operations.mdx` | `api-reference/variables/Gadgets` | The page describes these functions as gadgets but never linked their signatures (#1234) |

The intent follows the division of labour in #1208: the zkApp-context framing stays on Mina docs, the reference detail is on o1Labs. Nothing is copied and nothing is deleted — these are links only.

Two of the linked pages (`basic-types/arrays`, `basic-types/structs`, `basic-types/keypairs-and-signatures`) were not in the original comparison at all; they turned up when I diffed the live o1Labs sitemap against `docs/`.

All eleven URLs verified live (HTTP 200) on 2026-09-06. `static/llms-full.txt` regenerated.

### Not in this PR
The other 22 link-less pages. Some legitimately have no counterpart — `install-zkapp-cli.mdx` and `how-to-deploy-a-zkapp.mdx` are Mina-canonical by design — and the rest are covered by the stale-content issues (#1231, #1232) where a link is not the fix. #1234 tracks the sweep.

### Context
From the docs2 ↔ o1Labs overlap comparison indexed in #1237.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_012A5LmAdLpUcPELjgHg3C5e

PR #1242, opened 2026-09-06
ci: check outbound o1Labs documentation links and anchors

dkijania:
Part of #1235.

Nothing in CI verifies the links we point at `docs.o1labs.org`. There are **40** of them — 37 in `docs/` plus the `vercel.json` redirect destinations — and for the eleven guides delegated in Phase 2 of #1208, **those redirects are the only route a reader has to the content**. A silent break there is a dead end, not a detour.

### It reproduces both known breakages

Run against `main` today:

```
Checking 40 docs.o1labs.org references...

⚠️  o1Labs pages declare "documentation.o1labs.org" as their canonical host,
    but we link to "docs.o1labs.org". See MinaProtocol/docs2#1236.

❌ 2 broken reference(s):

  https://docs.o1labs.org/o1js/api-reference/namespaces/Experimental/functions/IndexedMerkleMap
    HTTP 404
    referenced from docs/zkapps/o1js/indexed-merkle-map.mdx:73

  https://docs.o1labs.org/o1js/zkapps/permissions#upgradeability-of-smart-contracts
    page is 200 but anchor #upgradeability-of-smart-contracts no longer exists
    referenced from docs/glossary.mdx:310
```

Found independently of the manual audit that produced #1235, which is the check I wanted before proposing it.

### Why anchors and not just status codes
The glossary case is the whole argument. The page returns 200, so a status-only checker sees nothing wrong, while the reader lands on the permissions page with no upgradeability section anywhere on it. #1238 fixes both of these; **this job will pass once that merges.**

### Why scheduled, not per-PR
These URLs are owned by `o1-labs/o1js-documentation-site`. They break through upstream edits, not through anything in a Mina docs pull request — blocking unrelated PRs on someone else's rename helps nobody. Weekly cron plus `workflow_dispatch`.

### The canonical-host warning
Every o1Labs page declares `rel="canonical"` pointing at `documentation.o1labs.org`, and their sitemap lists 1105 URLs all on that host — but it **does not resolve in DNS**, while `docs.o1labs.org` serves everything with a 200. We have 40 references pointing at the host their own site says is not canonical. The job surfaces this rather than failing on it, since it is a question for o1Labs (#1236), not a defect here.

### Shape
`scripts/check-o1labs-links.js`, matching the existing `validate-docker-images.js` style — plain CommonJS, `https` only, no new dependencies. Prettier-clean. One fetch per page, shared across anchors, 6-way concurrency.

https://claude.ai/code/session_012A5LmAdLpUcPELjgHg3C5e


.github/workflows/check-o1labs-links.yml
```diff
@@ -0,0 +1,27 @@
+name: Check o1Labs documentation links
+
+# Scheduled rather than per-PR on purpose: these URLs are owned by
+# o1-labs/o1js-documentation-site, so they break through upstream edits
+# rather than through anything in a Mina docs pull request. Blocking
+# unrelated PRs on someone else's rename helps nobody.
+
+on:
+  schedule:
+    - cron: '0 6 * * 1'
```
SanabriaRusso: I'd suggest to run them daily. This is a cheap job and we benefit by knowing in less that 24h if a link breaks due to edits upstream. @dkijania

SanabriaRusso (approved):

.github/workflows/check-o1labs-links.yml
dkijania: Agreed — changed to `cron: '0 6 * * *'` in c675f4d3. The job is around sixty HEAD requests, so a week's delay buys nothing.

While applying it I rebased the branch, which was 41 commits behind `main`. That mattered: run as it stood, the checker reported two broken references — the IndexedMerkleMap URL and the `#upgradeability-of-smart-contracts` anchor in the glossary — which are exactly the two that #1238 fixed after this branch was cut. Against current `main` it is clean:

```
Checking 60 docs.o1labs.org references...
⚠️  o1Labs pages declare "documentation.o1labs.org" as their canonical host, but we link to "docs.o1labs.org". See MinaProtocol/docs2#1236.
✅ All 60 references resolve.
```

The reference count also grew from 40 to 60 since this branch was written, because #1240 and #1241 added cross-links — which is the argument for the daily cadence getting stronger, not weaker.

PR #1248, opened 2026-09-14
ci: fail on any documented Docker image a reader cannot pull

dkijania:
`validate-docker-images` fails on `main`, so every PR that touches `docs/` inherits a red check — including #1244, which only adds a caution paragraph. This fixes the job and the references it was right to complain about.

## The rule

A reader must be able to pull every image the documentation shows. A private image is as useless to them as a missing one, so the job now fails on both:

- `❌ NOT FOUND (404)` — the registry has no such image.
- `❌ NOT PUBLIC (401)` — the registry refuses an anonymous pull.

No exception list. Hiding either case hides the problem.

## What was wrong

**The GCR check has never worked.** It sent an unauthenticated manifest request to `gcr.io/v2`. The Docker Registry HTTP API V2 answers that with `401` and an authentication challenge, which the script never answered, so *every* `gcr.io` image was reported as a soft error — public or not — and exited 0. `checkGcrImage` now requests an anonymous pull token and retries with it:

| image | before | after |
| --- | --- | --- |
| `gcr.io/distroless/static:latest` | ⚠️ 401 | ✅ 200 |
| `gcr.io/google-containers/pause:3.2` | ⚠️ 401 | ✅ 200 |
| `gcr.io/distroless/no-such-image-xyz:latest` | ⚠️ 401 | ❌ 404 |

The manifest request also accepts the manifest-list and OCI index media types, so a multi-architecture image is not misread as missing.

**The Berkeley Docker example could not work.** It ran `minaprotocol/mina-archive-migration:3.0.1-e848ecb-bullseye`. The *repository* returns 404 on Docker Hub — the public `minaprotocol` namespace holds 11 repositories and this is not one of them — and the migration tooling was removed from the `mina` repository after the upgrade. The page is deleted, `vercel.json` redirects it to the Debian example, and the installation page no longer names the image at all. The Debian packages are untouched and still resolve: `mina-archive-migration` is in `bullseye/stable` and `focal/stable` at `3.0.1-e848ecb` and `3.0.2-5f6848c`.

**Mesa Trail has no Docker images.** The pages pointed at `gcr.io/o1labs-192920`, whose token endpoint answers `DENIED: Unauthenticated request ... artifactregistry.repositories.downloadArtifacts`, and `4.0.0-rc2-3418329` was never pushed to Docker Hub. An operator could install `mina-devnet 4.0.0-rc2-3418329` from `packages.o1test.net` (channel `mesa-trail`, verified present) but could not get an image by any route. The Mesa Trail page now says so and keeps only the Debian install path — the Docker tab, the image list, and the Docker variants of the status and log commands are gone. The archive upgrade page says the same for the Mesa Trail version and marks the "extract the script from the Docker image" alternative as applying to the releases that publish one.

## Result

```
Total images found: 10
✅ Valid images: 10
❌ Invalid images: 0
⚠️  Errors: 0
```

Every image the docs still show is publicly pullable, and the check now proves it instead of skipping the ones it could not reach. If Mesa Trail images are published later, add them back and the check will confirm them.

The site builds clean (`npm run build`, exit 0) and `static/llms-full.txt` / `llms.txt` are regenerated.


scripts/validate-docker-images.js
```diff
@@ -154,6 +165,7 @@ async function checkDockerHubImage(imageName, tag) {
 
     return {
       exists: response.statusCode === 200,
+      requiresAuth: response.statusCode === 401 || response.statusCode === 403,
```
glyh: nit: this branch never fires in practice. The Docker Hub tags API answers 404 for private repos, not 401/403, so a private Docker Hub image is still reported as NOT FOUND. Harmless, just noting that NOT PUBLIC only applies to GCR.

glyh (approved):
LGTM. Verified: the validator passes locally (10/10), GCR anonymous token works for public repos and is DENIED for `o1labs-192920`, Mesa devnet/mainnet images exist on Docker Hub, and the migration tooling was removed from mina in 582abd4c1f. Nits below and inline.

nit: `docs/network-upgrades/mesa/mesa-trail.mdx:4`: the description still lists "Docker images" although the page is now Debian-only (the `docker` keyword on line 10 too). It is copied into `static/llms.txt`, so worth dropping and regenerating.

Issue #1249, opened 2026-09-14
epic: revive the zkApp tutorials — o1js 3.0.0, a project per tutorial, and CI that runs them


dkijania:
## Goal

Bring the zkApp tutorial series back to a state where **every tutorial is a project a reader can clone and run, on the o1js the reader will actually install, proved by CI on every pull request**. Then, and only then, discuss handing the series to o1Labs (`o1js.dev/docs` or `docs.o1labs.org/o1js/tutorials`), with tutorials that pass rather than tutorials that rot.

This is the umbrella for the work described in #1232, and it supersedes the whole tutorial-runner effort (see *Disposition* below).

## Why now

- **o1js 3.0.0 is the Mesa release.** It carries the Mesa transaction-limit constants, the 3.0 token contract permissions, the removal of `setFeeWithSnarkCost`, Cairo removal and native Kimchi. Mainnet forked to Mesa on 2026-09-03. The tutorials still declare `"o1js": "^2.0.0"` as a peer dependency — a range that can never resolve 3.x — and `04-zkapp-browser-ui/ui` declares `^1.*`, two majors behind.
- **The good news: the port looks cheap.** Copying each CI-exercised example to a scratch directory, setting the peer range to `^3.0.0` and installing `o1js@3.0.0`, all five build with **no source changes**:

  | example | `npm run build` on 3.0.0 |
  | --- | --- |
  | `01-hello-world` | ok — and `node build/src/main.js` produces the expected output, including the deliberate `assertEquals` failure demo |
  | `02-private-inputs-and-hash-functions` | ok |
  | `05-common-types-and-functions` | ok |
  | `09-recursion` | ok |
  | `10-account-updates` | ok |

- **Six tutorials plus the server-side guide have had no content commit since 2024** (#1232), and the three worst have no runnable project at all.
- **The machinery that was supposed to prevent this is gone.** `scripts/tutorial-runner.ts` was replaced by the examples matrix in `70915b36` (Oct 2024) and is referenced nowhere; sixteen issues and two PRs are still open against it.

## Current state

| Tutorial | Example project | CI treatment | Test files |
| --- | --- | --- | --- |
| 01-hello-world | `01-hello-world` | build + run | 0 |
| 02-private-inputs-hash-functions | `02-private-inputs-and-hash-functions` | build + run | 0 |
| 03-deploying-to-a-network | **none** | **none** | — |
[...]

PR #1252, opened 2026-09-15
ci: put the tutorials on o1js 3.0.0 and make CI run them

dkijania:
Phase 1 of #1249.

## The version gap

Every example declared a peer range that cannot resolve the o1js a reader installs today:

| package | was | now |
| --- | --- | --- |
| 01, 02, 05, 09, 10, 04/contracts, 07/contracts | `^2.0.0` | `^3.0.0` |
| `anonymous-message-board`, `interacting-with-zkApps-server-side` | `1.*` | `^3.0.0` |
| `feature-overview` | `^1.8.0` | `^3.0.0` |
| `04-zkapp-browser-ui/ui` | `^1.*` (dependency) | `^3.0.0` |

npm's `latest` is **3.0.0**, which is the **Mesa** release — Mesa transaction limits, 3.0 token contract permissions, `setFeeWithSnarkCost` removed, native Kimchi. Four packages were two majors behind, on a pre-fork library, and `feature-overview`'s suite — the only tests CI actually ran — was green on o1js 1.x. All lockfiles now resolve 3.0.0.

## One real breakage, fixed at the root

`feature-overview/src/upgradability/upgrade.test.ts` asserted four literal verification key hashes. A verification key is a function of the proof system as well as of the contract, so the Mesa proof system changes them — the test failed on 3.0.0.

Updating the constants would buy one release. The test now asserts what the upgradability example actually depends on:

- every contract compiles to a non-empty, non-zero verification key hash
- no two of `AddV1`, `AddV2`, `AddV3`, `AddV3Unsafe` share a hash — *"so an upgrade between them would not be observable"*

10 passing tests where there were 9. The rest of the file already compared against freshly compiled values and needed no change.

## The workflow now matches what each example can do

- **`run-examples`** — executes 01, 02, 05, 09 (`vote.js` + `rollup.js`) and 10 against a local blockchain
- **`test-examples`** — runs the suites of `feature-overview` and of the **04** and **07** contracts. Those two had real test files and were only ever built; their tests have never run in CI until now
- **`build-examples`** — the four with no offline entry point: the two front ends, the message board contract, and the server-side guide, which talks to `api.minascan.io/node/devnet` and cannot run deterministically in CI. Phases 2–3 of #1249 give these entry points
- **`o1js-latest-canary`** — builds all ten o1js packages against `o1js@latest` daily, `continue-on-error`, so the next release surfaces as a failing scheduled run instead of as a two-year-old tutorial

Plus: **runs on `pull_request`**, not only on push to main — tutorial breakage was previously only discoverable after merge — and the Node matrix moves **18/20/22 → 20/22/24**, dropping a release past end of life.

## Verified locally, all on o1js 3.0.0

```
01-main  PASS    09-vote    PASS    04-contracts-test  PASS
02-main  PASS    09-rollup  PASS    07-contracts-test  PASS
05-main  PASS                       feature-overview   10/10 pass
10-main  PASS
```

and the four build-only examples build, including the Next.js UI on o1js 3.0.0.


Shigoto-dev19 (approved):
Good to merge ✅

PR #1253, opened 2026-09-15
feat(tutorials): give custom tokens a runnable example project

dkijania:
Phase 2 of #1249 — the first of the three tutorials that have no project at all.

## What was wrong

Tutorial 8 links to four files under `examples/zkapps/08-custom-tokens/` — `BasicTokenContract.ts`, `main.ts`, `WhitelistedTokenContract.ts` and the `src/` directory. **None of them has ever existed.** The links point at `o1-labs/docs2` and 404 through the rename redirect.

Worse, the contract the tutorial teaches does not compile on current o1js. `SmartContract.token` was replaced by the `TokenContract` base class, so `this.token.mint(...)` and `this.token.send(...)` do not resolve on o1js 3.0.0 — a reader copying the tutorial gets a type error on the first build.

## What is here now

Everything runs on `Mina.LocalBlockchain` — no network, no faucet, no funded accounts, per the decision recorded in #1249.

**`BasicTokenContract`** — extends `TokenContract`, mints under an admin signature over `(amount, receiver)`, sends between accounts, and implements the `approveBase` that `TokenContract` leaves abstract:

```ts
async approveBase(forest: AccountUpdateForest) {
  this.checkZeroBalanceChange(forest);
}
```

so the approved updates must sum to zero and only `mint()` can raise the supply.

**`WhitelistedTokenContract`** — the "putting rules around a token" example the tutorial has always promised and never had. On-chain state is eight field elements and a public key costs two, so the contract stores a **Poseidon commitment** to the whitelist; methods take the list as an argument and check its hash against the commitment before checking membership.

**`main.ts`** — deploys, mints, transfers, prints balances and the amount in circulation at each step:

```
deploying the token contract...
total in circulation: 0
minting to alice...        total in circulation: 100000   alice balance: 100000
sending from alice to bob...  alice: 60000   bob: 40000
total in circulation is unchanged by a transfer: 100000
```

**19 tests, all passing in ~6s**, weighted towards the failure cases because those are the claims a reader is most likely to get wrong:

| suite | covers |
| --- | --- |
| `BasicTokenContract` (9) | symbol set on deploy, token id ≠ MINA token id, mint raises circulation, transfer does not, **forged admin signature rejected**, **signature bound to a different receiver rejected**, **overdraft rejected** |
| `WhitelistedTokenContract` (10) | commitment starts empty, **mint before any whitelist rejected**, whitelist accepted only under the zkApp key, **whitelist signed by another key rejected**, **mint to a non-member rejected**, **a list that does not match the commitment rejected**, **transfer to a non-member rejected** |

## Documentation

The tutorial's code blocks move to the current API, gain a paragraph explaining what `TokenContract` and `approveBase` are for, and their four links now point at this repository.

## CI

The example is wired into `run-examples` (executes `main.js`) and `test-examples` (runs the suite) on Node 22 and 24, plus the `o1js@latest` canary. `npm run build` for the site exits 0 and `llms-full.txt` is regenerated.


examples/zkapps/08-custom-tokens/src/BasicTokenContract.ts
```diff
@@ -0,0 +1,105 @@
+import {
+  AccountUpdateForest,
+  DeployArgs,
+  method,
+  Permissions,
+  PublicKey,
+  Signature,
+  SmartContract,
```
Shigoto-dev19: Nit: `SmartContract` is imported here but is never used.

examples/zkapps/08-custom-tokens/src/BasicTokenContract.test.ts
```diff
@@ -0,0 +1,166 @@
+import assert from 'node:assert';
+import { after, before, describe, it } from 'node:test';
```
Shigoto-dev19: **Nit:** `after` is imported from `node:test` but is never called.

Shigoto-dev19 (approved):
There are two small nits that I pointed that are related to unused imports; other than that the PR is good to merge.

Approving ✅

examples/zkapps/08-custom-tokens/src/BasicTokenContract.ts
dkijania: Good catch — dropped. `SmartContract` only appears in the class comment explaining that this contract extends `TokenContract` *rather than* `SmartContract`, so the import was never a value reference.

examples/zkapps/08-custom-tokens/src/BasicTokenContract.test.ts
dkijania: Dropped too. There is no teardown here — the local blockchain lives for the length of the process — so `after` was left over from the scaffold.

I swept the other five files in the example while I was at it; no other unused imports.

dkijania:
Rebased onto `main`, which had moved 24 commits including #1255. The conflict was one line in `test-tutorials.yml` — both this example and `03-deploying-to-a-network` claim a slot in the `test-examples` matrix — resolved by keeping both. Suite re-run after the rebase and the import fixes: **19/19 passing**.

dkijania:
Rebased again: #1254 merged after the previous rebase and claims slots in the same three matrices, so the same one-line conflict reappeared in `test-tutorials.yml` — three hunks this time (`run-examples`, `test-examples`, the canary), all resolved by keeping both entries. No duplicates in any job: run-examples 8, test-examples 6, build-examples 4, canary 13. Site build exits 0.

PR #864, opened 2024-02-23
:memo: Updated the deprecated code in tutorial and file

LuffySama-Dev:
Hi @barriebyron ,

I am creating this PR to resolve the issue number #756 .
The tutorial has multiple errors, I have asked in discord 
https://discord.com/channels/484437221055922177/1210453122376269935

Waiting for the resolution.
Meanwhile I have updated  the deprecated code.

Thank You 😃

Issue #756, opened 2023-11-27
HF audit zkApp tutorials: verify and run through code for Tutorial 6: Off-Chain Storage


barriebyron:
HF audit zkApp tutorials: verify and run through code for Tutorial 6: Off-Chain Storage

New issue to verify that the code runs for this tutorial as part of the tutorial audit
Earlier issues may not have satisfied the Definition of Done (DoD) where the code is run against the latest versions of o1js and zkApp CLI

looks like a community member is working on this https://github.com/o1-labs/docs2/pull/748

enderNakamoto:
The same issue is in the code example as well. There are more ssues in other tutorials as well, will send pull requests.

LuffySama-Dev:
Hey @enderNakamoto ,

Good Day!!
Are you still working on this issue ?
If not can I take a look at it ?

cc: @barriebyron 

Thank You 😃

barriebyron:
The more love we get on tutorials, the better they will serve our developers! Please check it out @LuffySama-Dev

enderNakamoto:
@LuffySama-Dev - it's all yours.

@barriebyron - there was supposed to be an overall audit of tutorials, is that still ongoing?

garwalsh:
Disable this until it is replaced

dkijania:
Done by #1254, merged today.

The audit asked for the tutorial 6 code to be verified and run through. It is now verified continuously rather than once: `examples/zkapps/06-offchain-storage` runs in `run-examples` and its 9 tests run in `test-examples`, on Node 22 and 24, on every pull request, plus a daily build against `o1js@latest`.

The tests cover the two properties the page claims that are easy to get wrong — a write is invisible until it is settled, and a settlement drops an update that names a stale value instead of overwriting the newer one.

Issue #695, opened 2023-10-27
Tutorial 6 off-chain storage doesn't work


qwadratic:
[Tutorial 6](https://docs.minaprotocol.com/zkapps/tutorials/offchain-storage): 
- The code for the tutorial comes with a different project structure which the tutorial does not refer to. Also in the `06-offchain-storage` folder, there is `ui/pages` directory which seems to be added by mistake
- The `contracts/src` directory has `SignedMessageBoard` which is not covered in the tutorial. Anyway, I am not able to run it (see the error message below) and unfortunately, I can’t find a way to fix it.
```
/Users/ii/mina/docs2/examples/zkapps/06-offchain-storage/offchain-storage-zkapp/contracts/node_modules/o1js/dist/node/bindings/compiled/_node_bindings/snarky_js_node.bc.cjs:7467
         throw err;
         ^

Error: "Assert_failure src/lib/transaction_logic/zkapp_command_logic.ml:1286:17"
    at Object.sendTransaction (o1js/dist/node/lib/mina.js:249:24)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async sendTransaction (o1js/dist/node/lib/mina.js:745:12)
    at async Object.send (o1js/dist/node/lib/mina.js:176:24)
    at async main (file:///Users/ii/mina/docs2/examples/zkapps/06-offchain-storage/offchain-storage-zkapp/contracts/build/src/messageBoardMain.js:29:5)
```
- For some reason, the tutorial asks the user to deploy the app to Berkeley, however, in other tutorials, we are happy to have local deployment.
- The correct command to run the off-chain storage server should be `node offchain-storage-zkapp/contracts/node_modules/experimental-offchain-zkapp-storage/build/src/server/storageServer.js`
- For some reason when I do “zk deploy” I get the following error
```
Failed to send transactions to the relayer. Errors: Couldn't send zkApp command: (Verification_failed
 "Invalid_signature: [B62qmVz7pPiLXPvz2nPkuK3K5akjrePAVtdBVMfeyixrccgqKTQte8K]") 
```
however other contracts are deploying successfully with the same feePayer account.
- As a result, I am unable to run the tutorial

dkijania:
Fixed by #1254, merged today.

The complaint was that the tutorial's project structure did not match the code it pointed at. Both are now real and they match: `examples/zkapps/06-offchain-storage` exists — it never did before — and the page is built around it.

[...]

PR #1254, opened 2026-09-15
feat(tutorials): rebuild offchain storage on the o1js offchain state API

dkijania:
Phase 2 of #1249. Second of the three tutorials with no project. Approach chosen in [this thread](https://github.com/MinaProtocol/docs2/issues/1249).

## What tutorial 6 was

It taught `experimental-zkapp-offchain-storage`, an npm package **last published 2022-11-17**, whose dependencies are `express@4`, `cors@2.8` and `xmlhttprequest@1.8`. The page pins **o1js 0.12.1** and **zkApp CLI 0.11.2**, tells the reader to start a REST server out of `node_modules`, and links to four example files that have never existed in this repository. It carries a grant section soliciting work on that library.

It is also the tutorial behind #695 ("Tutorial 6 off-chain storage doesn't work") and #756, and behind #864 — a community fix from 2024 that rotted in review until the directory it patched no longer existed.

## What it teaches now

o1js ships the same idea in-house, so the page teaches `Experimental.OffchainState`: state declared as fields and maps, data carried in **actions** so it is available to anyone reading the chain, a **commitment** in on-chain state, and a **settlement proof** folding pending actions into that commitment.

```ts
export const offchainState = OffchainState(
  {
    numbers: OffchainState.Map(Field, Field),
    total: OffchainState.Field(UInt64),
  },
  { logTotalCapacity: 10, maxActionsPerUpdate: 4 }
);
```

The map takes the place of the Merkle tree the old page kept on a server, and the settlement proof takes the place of trusting whoever ran it.

## The example

Runs on `Mina.LocalBlockchain` — no server, no network, no funded accounts.

`npm start`:

```
writing 42 at index 1...
before settling, index 1 is: not set yet
settling...
after settling, index 1 is: 42
entries written: 1
updating index 1 from 42 to 43...
index 1 is now: 43
```

**9 tests**, weighted to the two properties a reader is most likely to get wrong:

- a write is **invisible until it is settled** — the test asserts `isSome === false` immediately after the write transaction
- a settlement **drops an update that names a stale value** instead of clobbering the newer one: after index 1 moves 42 → 43, an update still expecting 42 leaves 43 in place
- and: a `setNumber` on an occupied index is dropped, a second entry does not disturb the first, the counter tracks entries, an untouched index stays empty

That second one is the argument for this API over hand-rolling a tree, so it is worth a test rather than a sentence.

## Prose

The framing on storage and decentralization is kept. Gone: the storage-server steps, the o1js 0.12.1 / CLI 0.11.2 pins, the four dead links, and the grant section soliciting work on the retired library. Added: what settlement is for, and the difference between `update` and `overwrite`.

## CI

`run-examples` executes `main.js`, `test-examples` runs the suite, on Node 22 and 24, plus the `o1js@latest` canary. Site build exits 0, `llms-full.txt` regenerated.

Note: this branches from `main`, so it will conflict trivially with #1253 in `test-tutorials.yml` — one matrix line each.


Shigoto-dev19 (approved):
LGTM!

dkijania:
Rebased onto `main`. Same one-line conflict in `test-tutorials.yml` as #1253 — this example and `03-deploying-to-a-network` both adding a `test-examples` entry — resolved by keeping both. No duplicate matrix entries in any of the four jobs, `llms-full.txt` regenerated against current `main`, and the site build exits 0.

PR #1257, opened 2026-09-16
docs: point the GraphQL examples at the o1Labs devnet endpoints

### LINKED ISSUE #840: Graph QL endpoint in tutorials
I have noticed that the graph ql from the tutorials would not work.

I have researched web with Google and added a number of developer resources and links in my post on https://dspyt.com/Mina-protocol

https://mina-berkeley-graphql.aurowallet.com/graphql

worked for me on Berkeley

dkijania:
Closes #840.

## The measurement

Every tutorial that talks to a node sent the reader to `api.minascan.io`. From one machine, repeatedly:

| endpoint | who runs it | result |
| --- | --- | --- |
| `devnet-plain-1.gcp.o1test.net/graphql` | o1Labs | **5/5 · 200 · ~0.13s** |
| `devnet-archive-node-api.gcp.o1test.net` | o1Labs | **200 · 0.23s** |
| `api.minascan.io/node/devnet/v1/graphql` | Minascan | **2/6**, 0.17s when it answered, 15–25s timeout when it did not |
| `api.minascan.io/archive/devnet/v1/graphql` | Minascan | **0/3** |

Not down — flaky. Which is exactly what #840 reported in **February 2024**: *"the graph ql from the tutorials would not work."* The report stayed true for two years because nothing checked it.

## The change

**13 references across 8 pages and 2 examples** move to the o1Labs endpoints: tutorials 3 and 4, the server-side guide, the Angular guide, `fetch-events-and-actions`, and three `introduction-to-zkapps` pages, plus `04-zkapp-browser-ui/ui/app/zkappWorker.ts` and `interacting-with-zkApps-server-side/src/main.ts`.

**Links to the Minascan explorer are untouched** — `minascan.io/devnet/tx/...` is for a human to look at a transaction, not for a client to call. Only API endpoints moved.

## A third dead endpoint, found by the new check

The Berkeley appendix still named `https://qanet.minaprotocol.network/graphql`. **That host no longer resolves at all.** The page already tells the reader to substitute their own endpoint, so it now names one that answers.

## The part that stops this recurring

`scripts/check-graphql-endpoints.js` finds every GraphQL endpoint referenced from `docs/` and `examples/`, sends each the smallest query any GraphQL server answers (`{ __typename }`), and **retries three times** so that one timeout is noise rather than news. Placeholders (`example.com`, `your-mina-node`) are skipped.

It runs as a second job in the daily `check-o1labs-links` workflow, so an endpoint that stops answering fails a scheduled job the next morning instead of sitting in an issue for two years. `npm run check-graphql-endpoints` runs it locally.

```
Checking 2 GraphQL endpoint(s)...
✅ https://devnet-plain-1.gcp.o1test.net/graphql
✅ https://devnet-archive-node-api.gcp.o1test.net
✅ All 2 endpoint(s) answer.
```

Site build exits 0.


docs/zkapps/tutorials/interacting-with-zkapps-server-side.mdx
Shigoto-dev19: This line still says the example uses `minascan.io`, but the code above it was just changed to a different URL. Worth updating this sentence too so it matches.

scripts/check-graphql-endpoints.js
Shigoto-dev19, line 1: In line 372, the endpoints are checked one after another instead of at the same time. Not a problem now with just 2 endpoints, but worth switching to run them in parallel once more get added, so the check doesn't get slower over time.

Shigoto-dev19 (changes requested):
Looks good overall!

Before merging, two things need a fix:
1. One example project (config.json) still points to the old broken link that this PR is meant to remove.
    - `examples/zkapps/interacting-with-zkApps-server-side/config.json` still points to the old `minascan.io` endpoint (the flaky one this PR is fixing). It looks like it was missed. Also, the new checker script won't catch this in the future because it doesn't scan `.json` files.
 
2. One paragraph of text still mentions the old service, even though the code sample right above it was already updated.

Once the two items above are fixed, this is good to merge.

docs/zkapps/tutorials/interacting-with-zkapps-server-side.mdx
dkijania: Fixed in f10796a8 — the sentence now names `devnet-plain-1.gcp.o1test.net`, matching the sample above it. `llms-full.txt` was regenerated so `check-llms-txt` stays green.

scripts/check-graphql-endpoints.js
dkijania: Done in f10796a8 — the probes now run under a single `Promise.all` instead of one after another. A full run goes from 22s to 2s, and adding an endpoint no longer costs wall-clock time.

Results are still collected in discovery order rather than completion order, so the output does not depend on which host answers first:

```js
const entries = [...endpoints];
const results = await Promise.all(entries.map(([url]) => check(url)));

for (const [index, [url, files]] of entries.entries()) {
  const result = results[index];
  ...
}
```

The 3-attempt retry is unchanged, so a flaky host still gets its second and third chance — those attempts are just no longer serialised behind other endpoints.

dkijania:
Thanks — both were real, and the first one was the more serious of the two. All three items are addressed in f10796a8.

### 1. `config.json` still pointing at minascan

You are right that this was the one that mattered. `examples/zkapps/interacting-with-zkApps-server-side/config.json` is what `zk deploy` actually talks to, so the example would have kept hitting the flaky endpoint no matter what the page said. It now uses `https://devnet-plain-1.gcp.o1test.net/graphql`, matching `src/main.ts`.

You were also right about *why* it was missed: the checker only read `.mdx`, `.md`, `.ts` and `.tsx`. It now reads `.json` as well, so a config file cannot drift out of the check again. I confirmed the new behaviour by putting a dead URL back into `config.json` — it showed up in the run — then restored the file.

One note on scope: I first added `.js`/`.jsx` too, and backed it out. It picked up a **committed** minified Next.js chunk under `examples/zkapps/04-zkapp-browser-ui/ui/out/`, which still has the old minascan URL baked in. That bundle is generated output rather than a source file, so failing CI on it would be noise. Worth flagging separately though: it is stale relative to its source and will stay stale until that UI is rebuilt.

### 2. The paragraph still naming the old service

Fixed — `interacting-with-zkapps-server-side.mdx:95` now names `devnet-plain-1.gcp.o1test.net`. `llms-full.txt` regenerated so `check-llms-txt` stays green.

### 3. Sequential probes

Switched to `Promise.all`. A full run drops from 22s to 2s and no longer gets slower per endpoint added. Output order is still discovery order, not completion order, so the log does not depend on which host answers first. The 3-attempt retry is unchanged.

### Verification

- Checker: 2 endpoints found, both answer.
- `npm run build`: exit 0.
- CI on this push: 49 green, 6 still running, 0 failures.

Ready for another look.

Shigoto-dev19 (approved):

PR #1256, opened 2026-09-16
fix: put tutorials 6 and 8 back in the sidebar and in llms.txt

dkijania:
Part of #1249. Three pre-existing navigation defects, found while auditing the tutorial numbering.

**The numbering itself is sound** — files run 01–10, every page's title matches its filename, the two unnumbered pages are deliberate, and 11/12 exist only inside stale PRs. What was broken is everything around it.

## 1. The sidebar listed ten of twelve tutorials

`offchain-storage` (6) and `custom-tokens` (8) were absent. Docusaurus derives previous/next from the sidebar, so those two pages had **no navigation entry and no prev/next link** — reachable only by knowing the URL.

`git log -S` shows how they went: both were dropped by the automated *"Build and publish new o1js API reference docs"* commits — `custom-tokens` on **2024-05-15**, `offchain-storage` on **2024-06-27** — when `scripts/update-sidebars-o1js-api.js` regenerated the file and lost them.

They have been unreachable for over two years. That is very likely *why* two of the three tutorials with no example project (#1253, #1254) were those two: nobody walked past them.

## 2. The next-tutorial chain ended at 7

| | before | after |
| --- | --- | --- |
| 5 → 6 | ✅ already linked | unchanged |
| 6 → 7 | ✅ | unchanged |
| **7 → 8** | ❌ page ended with its example link and nothing to follow | **links to 8** |
| 8 → 9 → 10 | ✅ | unchanged |
| **10 → end** | ❌ nothing | **names the two unnumbered walk-throughs** |

Tutorial 7's example link also pointed at `o1-labs/docs2`; it now points here.

## 3. `llms.txt` listed two of twelve tutorials

The AI-facing index contained only `anonymous-message-board` and `interacting-with-zkapps-server-side` — **all ten numbered tutorials were missing**.

`scripts/generate-llms-index.mjs` resolved a sidebar id straight to `docs/<id>.mdx`, so `zkapps/tutorials/hello-world` looked for `hello-world.mdx` and never found `01-hello-world.mdx`. The site strips a number prefix when deriving an id; the script did not, and the failure was swallowed by a `catch`. It now resolves both spellings.

```
before: 2 tutorials, 129 pages, 25.9 KB
after: 12 tutorials, 139 pages, 27.5 KB
```

[...]

PR #1258, opened 2026-09-16
docs: add a tutorial on upgrading a zkApp

dkijania:
Phase 6.1 of #1249 — the cheapest tutorial on the board, because the code already exists and already passes.

## Why this one first

`examples/zkapps/feature-overview/src/upgradability/` has held `AddV1`, `AddV2`, `AddV3` and `AddV3Unsafe` with a passing suite for a long time, built and tested on every pull request since #1252. **No page in these docs teaches it.** o1Labs documents upgradability as a concept but ships no worked project, so nothing showed a reader what an upgrade actually does to state.

Only the page was missing. This adds it, as an unnumbered tutorial beside the anonymous message board, leaving 11 and 12 free for the restorations planned in #1249.

## What it teaches

**The upgrade that works.** Deploy `AddV1` (`num` starts at 1, `update()` adds 2), then replace the verification key with `AddV2`'s — a one-number change that compiles to a different key — through a signed account update. Call `update()` again: `num` is 7. The state survives; only the code reading it changed.

**Adding state.** `AddV3` declares `callCount` *after* `num`, upgrade, call `add10()`: `num` = 17, `callCount` = 1. The old value is still where the new contract expects it.

**The same upgrade done unsafely.** `AddV3Unsafe` is identical except the two fields are declared the other way round. On-chain state is positional — eight field elements in a fixed order, with the names living only in your TypeScript:

| slot | V1 reads it as | V3Unsafe reads it as |
| --- | --- | --- |
| 0 | `num` = 3 | `callCount` = 3 |
| 1 | unused = 0 | `num` = 0 |

Call `add5()` and the account holds `callCount` = 4 and `num` = 5. **Nothing fails** — no error, no rejected transaction, no warning. The balance became a call counter. That is the lesson the page exists for: add new state fields after the existing ones, never reorder them.

Every number above comes from assertions in the existing test suite, so the page cannot drift from the code without CI noticing.

## One thing I got wrong and fixed

My first draft named the permissions `impossible()` and `proof()`. The real API is `Permissions.VerificationKey.{ signature, impossibleDuringCurrentVersion, proofDuringCurrentVersion, none }`, and those two names are not cosmetic: **after a hard fork that increments the transaction version, both revert to `signature`**, so the original owner can redeploy. An immutable zkApp is immutable only until the protocol changes under it — worth a reader knowing before they choose.

Site build exits 0. Conflicts with #1256 on one `sidebars.js` line, since both add tutorial entries.


Shigoto-dev19 (approved):
Needs a rebase before it can merge.

Other than that, approving ✅

PR #1259, opened 2026-09-16
feat(examples): make the message board runnable and tested

### LINKED ISSUE #58: Add deploy() function to the Anonymous Message Board tutorial
> burnto says:
> Currently if you follow the step-by-step message board tutorial, it omits discussion of deployment and implementing a deploy function, which is a pretty critical part of the development cycle.
From an issue submitted to the old docs repo: https://github.com/MinaProtocol/docs/issues/232

Tagging @jackryanservia to consider this.

dkijania:
Phase 3 and 6.5 of #1249. Closes #58.

## The gap

`examples/zkapps/anonymous-message-board` shipped a **configured jest harness with zero test files** and no entry point, so CI could only build it. It is one of nine packages in that state — this is the first.

## What it does now

**`npm start`** deploys to a local blockchain and runs the flow — no network, no faucet, no funded accounts:

```
deploying the message board...
approved posters:
  user1: B62qkCkKQ5vSRkTQyEEj22vnHH4hVKfp4pc3QYd4GPSv7u6S4aBE4Fx
Bob publishes a message...       message: 42
SuperBob publishes another...    message: 99
history hash is the chain of both messages: true
Jack tries to publish...         rejected, as it should be
```

**`npm test`** asserts the same ground in 8 tests, weighted to the two claims worth proving rather than describing:

- **the history hash chains** — `Poseidon.hash([m3, Poseidon.hash([m2, Poseidon.hash([m1, 0])])])`, so the on-chain value commits to every message in sequence and one cannot be dropped unnoticed
- **an unknown key is refused** — both Jack's, which the contract was never given, and a freshly generated one

## A defect running it surfaced

`init()` never called `super.init()`, so **every deployment printed an o1js warning**:

> the `init()` method was called without overwriting the entire state. This means that your zkApp will lack the `provedState === true` status which certifies that the current state was verifiably produced by proofs (and not arbitrarily set by the zkApp developer).

Both the contract and the tutorial section that teaches `init()` now start with `super.init()`, with a sentence explaining what it buys. The warning is gone.

Worth noting what I did **not** do: I first wrote a test asserting `provedState === true`, and it failed. That status can only become true once state has been produced by proofs, and this suite runs with `proofsEnabled: false` so it finishes in 12 seconds. Rather than weaken the assertion or slow the suite to nothing, the test is gone and a comment in its place records why.

## #58, filed 2022-10-07

*"Add deploy() function to the Anonymous Message Board tutorial."* The page now has a **Deploy and run it** section: how to deploy, what `npm start` prints, and the deployment transaction itself, with the account creation fee, the verification key and `init()` called out as the same mechanics as tutorial 3.

## CI

The example moves out of `build-examples` into `run-examples` and `test-examples` on Node 22 and 24, and stays in the `o1js@latest` canary. No duplicate matrix entries: run 9, test 7, build 3, canary 13.

Site build exits 0.


Shigoto-dev19 (approved):
Good to merge ✅

